PR #291, opened 2023-06-03
fix: message canvas 0 height false positives

### LINKED ISSUE #263: Message of canvas 0 height error
**Describe the bug**
Could be the position absolute? or maybe conflict with BulmaCSS.

**Steps**
1. Go to https://serene-mousse-562782.netlify.app/mixHtml
2. Open the console

**Expected behavior**
Clearly the canvas is not 0 height, in fact has window-size property, 

**Screenshots**
image

image

**Additional context**
Please ignore the Color warning, that's my mistake

JaimeTorrealba:
I added a timeout is the fastest way to solve this issue, but also I changed the DOM element to the current canvas (not to the container) 
This because you can use something like
canvas{
position: absolute
height: 100vh
...
}
And render the scene but, the log appears because the container has 0 height

let me know

close #263

### DIFF
--- a/src/composables/useRenderer/index.ts
+++ b/src/composables/useRenderer/index.ts
@@ -158,10 +158,12 @@ export function useRenderer(options: UseRendererOptions) {
   const { pause, resume } = useRenderLoop()
   const aspectRatio = computed(() => width.value / height.value)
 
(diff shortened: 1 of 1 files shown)


alvarosabu (approved):

alvarosabu:
Thanks for the fix @JaimeTorrealba

Issue #282, opened 2023-05-24
Raycaster does not work properly when scene is not in full screen


Tinoooo:
### Describe the bug

The click-event on meshes does not properly work when the scene is not shown full screen.

### Reproduction

https://stackblitz.com/edit/tresjs-basic-hjvh64?file=src%2Fcomponents%2FTheExperience.vue

### Used Package Manager

pnpm

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/Tresjs/tres/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/Tresjs/tres/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://tresjs.org/guide).
- [X] Check that there isn't [already an issue](https://github.com/tresjs/tres/issues) that reports the same bug to avoid creating a duplicate.
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #304, opened 2023-06-16
fix: raycaster does not work properly when scene is not in full screen

Tinoooo:
# Fixes
* #282
* `pointerMove` does now trigger only when the mouse actually moves, not on every `requestAnimationFrame` callback.
* The `click` event behaves now just like in the DOM. That is only if `pointerup` fired on the same element as `pointerdown` did before. The click event used to fire in the same cases as `pointerup` would fire.

# Breaking Changes
* The `pointerLeave` event does not provide an intersection object, since leaving means no more intersection. 😄


alvarosabu (approved):
Thanks a lot, @Tinoooo amazing work 👏🏻

PR #306, opened 2023-06-17
fix: Remounting canvas

### LINKED ISSUE #302: Nuxt3 + Tresjs rendering issue(when remounting, not displaying)
### Describe the bug

Hi, I'm using TresJs v2(@tres/core) in Nuxt3.
Through your docs and kind explanations, I succeeded in configuring the environment and displaying 3d objects in Nuxt3.

However, if a vue component (I made) containing the TresCanvas tag is unmounted according to the page route movement or certain conditions, and then re-enters the page and is newly mounted, it seems that the object is not rendered on the canvas tag.

Can you check this?
You can see an example in the reproduction link below. You can check it when you toggle v-if with the toggle button.
Thank you.

### Reproduction

https://stackblitz.com/edit/nuxt-starter-36xfsn?file=nuxt.config.ts,app.vue (in Nuxt3)
https://stackblitz.com/edit/tresjs-basic-i4h4kk?file=src%2FApp.vue (same in Vue3)

### Steps to reproduce

_No response_

### System Info

```shell
npm packages version
- nuxt: 3.5.3
- three: 0.153.0
- @tresjs/core: 2.1.3
```

### Used Package Manager

yarn

### Code of Conduct

cjboy76:
Close #302 
Appreciate you for your incredible work.
I found that while remounting the canvas,  the perspective camera isn't rendered so that it causes this issue.

Also the condition with `firstCamera` is not needed because `setFirstCamera` only pushes camera if there is no cameras in state, perhaps there is some reasons why doing this.
Let me know any further updates.

### DIFF
--- a/src/core/nodeOps.ts
+++ b/src/core/nodeOps.ts
@@ -22,7 +22,6 @@ const OBJECT_3D_USER_DATA_KEYS = {
 
 const { logError } = useLogger()
 
-let firstCamera = true
 export const nodeOps: RendererOptions<TresObject, TresObject> = {
   createElement(tag, _isSVG, _anchor, props) {
     if (!props) props = {}
@@ -47,8 +46,7 @@ export const nodeOps: RendererOptions<TresObject, TresObject> = {
       }
       instance = new target(...props.args)
     }
-
-    if (instance.isCamera && firstCamera) {
+    if (instance.isCamera) {
       if (!props?.position) {
         instance.position.set(3, 3, 3)
       }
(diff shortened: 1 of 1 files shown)


alvarosabu:
Hi @cjboy76 first of all thanks for taking the time to create this PR. The solution does fix the issue of remounting but unfortunately breaks the HMR.

Although this is our fault because we recently figured out that using Composables like `useCamera` that access the provider context state **shouldn't** live inside the custom renderer `nodeOpts` (Composables need a component setup context which is lost constantly).

I'm gonna refactor the camera logic on #307, hope you don't mind if I close this PR.

Screenshot 2023-06-19 at 08 25 26

cjboy76:
@alvarosabu No problem, thanks for your reply 👍

alvarosabu:
@cjboy76 fixed here https://github.com/Tresjs/tres/pull/308

PR #315, opened 2023-06-22
feat: export utility functions

Tinoooo:


src/index.ts
```diff
@@ -5,6 +5,8 @@ export * from './core/catalogue'
 export * from './components'
 export * from './types'
 
+import { normalizeColor } from './utils/normalize'
```
alvarosabu, line 8: @Tinoooo now that we are here, wouldn't be better to export them all?

alvarosabu (changes requested):

alvarosabu (approved):

PR #320, opened 2023-06-26
docs: fix error in example code

nirazul:
There was an obvious error in one of the examples that is fixed in this PR.
Just a simple copypaste error probably :)


alvarosabu:
@nirazul thanks a lot for the fix 🙏🏻

PR #323, opened 2023-06-28
chore: refactor components to sfc

alvarosabu:
Check #324 first


src/components/TresCanvas.vue
```diff
@@ -0,0 +1,35 @@
...
+}
+
+defineProps<TresCanvasProps>()
+
+const tres = useTresProvider()
+
+defineExpose(tres)
+</script>
+<template>
+<TresScene v-bind="$props">
```
Tinoooo, line 32: Typescript gives me an error here (Argument of type .... not assignable ...)

Tinoooo (approved):

src/components/TresCanvas.vue
alvarosabu: Might be local, otherwise the build would fail

PR #333, opened 2023-07-04
feat: 331 new context for state

### LINKED ISSUE #331: New and better state context provider
### Description

Based on https://github.com/orgs/Tresjs/discussions/321

As a plugin author, I will like to have a reactive state with the following characteristics:

- Performant, (use of `shallowRef` where needed)
- read-only
- Strictly Typed
- Reduce side effects on watchers
- Needs to work on cientos and the rest of ecosystem via `provide/inject` 

### Suggested solution

| Name                               | Type                 | Requirement | Details                                                                                                                                                             |
| ---------------------------------- | -------------------- | ----------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| scene                              | readonly, shallowRef | essential   |                                                                                                                                                                     |
| cameras                            | readonly, shallowRef | essential   | userData.TRES_IS_ACTIVE_CAMERA = true<br>const cameras = shallowRef([]) return { cameras: readonly(cameras)}                                                        |
| camera                             | computed, readonly   | essential   |                                                                                                                                                                     |
| pushCamera(newCamera: Camera, setActive: boolean) | function | essential | cameras should be watched and removed from the cameras array if they got removed from the scene |
| removeCamera(camera: Camera) | function | essential ||
| setCamera(cameraId string) | function | essential ||
| renderer                           | readonly, shallowRef | essential   | setRenderer(renderer: Renderer)<br>maybe only callable once?                                                                                                        |
| setRenderer(renderer: Renderer) | function | essential ||
| sizes (height, width, aspectRatio) | readonly             | handy       | sizes: {<br>&nbsp;&nbsp;&nbsp;&nbsp;height: Ref<number><br>&nbsp;&nbsp;&nbsp;&nbsp;width: Ref<number><br>&nbsp;&nbsp;&nbsp;&nbsp;aspectRatio: Computed<number><br>} |

### Alternative

_No response_

### Additional context

_No response_

### Validations

alvarosabu:
Closes #331


alvarosabu:
@Tinoooo I added an alias for `useTresContext` called `useTres` to:
- Fix the issue with the docs build
- Avoid having a BREAKING CHANGE that may confuse the user and bump the version to 3.0

Tinoooo:
> @Tinoooo I added an alias for `useTresContext` called `useTres` to:
> 
> * Fix the issue with the docs build
> * Avoid having a BREAKING CHANGE that may confuse the user and would have to bump the version to 3.0 instead of 2.5 which is the one we want.

I like that. But the breaking change exists anyways, because the `useTres` now returns Refs instead of the objects themselves.

Tinoooo (approved):

PR #344, opened 2023-07-11
fix: fake a `VNodeRef` on types to accept strings and null

### LINKED ISSUE #301: Typescript compiler error for refs with tresjs
### Describe the bug

I am using tresjs and would like to reference my newly loaded .glb  and a directional light.
I should say, the page actually functions as expected, however the red squiggly line appears in my local VSCode with an error, and so would like to nail it down if possible.

Final note: this typescript compiler error only occurs on my local VSCode system, as can be seen from the StackBlitz link below, it doesn't occur there.
So I'm not sure if this is even a bug, but I'd like to ask for help to remove the red squiggly line from my local system, thanks if possible.

### Reproduction

https://stackblitz.com/edit/vitejs-vite-qn7w4v

### Steps to reproduce

I've tried changing the type of ref to a few things, eg: <THREE.DirectionalLight>, however nothing seems to remove this error:

(property) ref?: ((Ref<THREE.DirectionalLight> | Readonly<Ref<THREE.DirectionalLight>>) & VNodeRef) | undefined

### System Info

```shell
Here is my package.json for that project:

{
  "name": "tresjs_1",
  "private": true,
  "version": "0.0.0",
  "type": "module",
  "scripts": {
    "dev": "vite",
    "build": "vue-tsc && vite build",
    "preview": "vite preview"
  },
  "dependencies": {
    "@tresjs/cientos": "^2.1.3",
    "@tresjs/core": "^2.1.2",
    
    "three": "^0.153.0",
    "three-stdlib": "^2.23.9",
    "vue": "^3.2.47"
  },
  "devDependencies": {
    
    "@types/three": "^0.152.1",
    "@vitejs/plugin-vue": "^4.1.0",
    "typescript": "^5.0.2",
    "vite": "^4.3.9",
    "vue-tsc": "^1.4.2"
  }
}
```

### Used Package Manager

npm

### Code of Conduct

alvarosabu:
Closes #301

### DIFF
--- a/src/types/index.ts
+++ b/src/types/index.ts
@@ -1,6 +1,6 @@
 /* eslint-disable @typescript-eslint/ban-types */
 import type * as THREE from 'three'
-import { DefineComponent, Ref, VNode } from 'vue'
+import { DefineComponent, Ref, VNode, VNodeRef } from 'vue'
 
 // Based on React Three Fiber types by Pmndrs
 // https://github.com/pmndrs/react-three-fiber/blob/v9/packages/fiber/src/three-types.ts
@@ -144,7 +144,7 @@ type EventProps<P> = P extends RaycastableRepresentation ? Partial<EventHandlers
 
 export interface VueProps<P> {
   children?: VNode[]
-  ref?: Ref<P>
+  ref?: string | null | Ref<P>
   key?: string
 }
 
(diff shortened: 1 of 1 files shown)


src/types/index.ts
```diff
@@ -1,6 +1,6 @@
 /* eslint-disable @typescript-eslint/ban-types */
 import type * as THREE from 'three'
-import { DefineComponent, Ref, VNode } from 'vue'
+import { DefineComponent, Ref, VNode, VNodeRef } from 'vue'
```
Tinoooo, line 3:
```suggestion
import { DefineComponent, Ref, VNode } from 'vue'
```

Github Actions says it is not used 🤷‍♂️

PR #345, opened 2023-07-11
feat: add spanish docs

### LINKED ISSUE #123: Add i18n to Tresjs documentation
# description
 
To expand the project, it will be necessary to add translations in different languages, for this we need to create the base.
Currently there are intentions to add Spanish and Chinese.

## To discuss

We have several strategies to achieve this goal, but here I will expose two of them.

### 1. Using the current feature provide by vitepress (check it out)[https://vitepress.vuejs.org/guide/i18n] 

1. We need to make the redirection manually (explained in the link) through netlify
2. We need to reorganize the folder structure of the Docs 
3. Vitepress will make the rest for us
4. Currently I have found no projects using this feature

### 2. Using the I18n plugin

1. We need to install and setup the library
2. We don't need to update folders but we'll need to update the content, with the current JSON
3. We need to create the navbar options manually
4 Be aware of diferent formats like plurals, dates etc.
5 Implies more work to do, but it is a tested solution

JaimeTorrealba:
close #123


JaimeTorrealba:
Vitepress doesn't handle for you. Please check https://vitepress.dev/guide/i18n#separate-directory-for-each-locale. For more info.

Once this is ready I could compromise to update Spanish docs, each time English docs changes (only create an issue will be necessary).

But as I have done this translation twice (or three time) I believe you need to finish it :)

JaimeTorrealba:
If you check, I already added that lines into the vite.config

image

So everything should be ready for netlify. You (and only you) can try in local if you have the netlify CLI. 

@alvarosabu as I said, I did updated the translations twice (or three times). So I don't want to do it again, you have to decide if you want to do it again from 0 or continue this one. I'm not going to translate again. I compromise when everything is ready I maintain the spanish docs

PR #364, opened 2023-08-12
feat: use render instead createApp

enpitsuLin:
just use the render function instead of creating a Vue app instance just work

avoid vue devtools crash
image

***

I also have an idea about Vue devtools, how about creating a plugin with TresJS to improve DX of that?

I have taken a little time to check the Vue devtools doc, it seems we can display something in THREE scenes like object list, camera objects, and light objects. etc. just like Vue Router did

### DIFF
--- a/src/components/TresCanvas.vue
+++ b/src/components/TresCanvas.vue
@@ -7,7 +7,7 @@ import {
     type ShadowMapType,
     type ToneMapping,
 } from 'three'
-import { computed, h, onMounted, ref, shallowRef, watch, watchEffect } from 'vue'
+import { Fragment, computed, defineComponent, h, onMounted, provide, ref, shallowRef, watch, watchEffect } from 'vue'
 import { useTresContextProvider } from '../composables'
 import { render } from '../core/renderer'
 
@@ -17,7 +17,6 @@ import {
     useRenderLoop,
 } from '../composables'
 
-import { Fragment } from 'vue'
 import type { RendererPresetsType } from '../composables/useRenderer/const'
 import type { TresCamera, TresObject } from '../types/'
 
@@ -80,7 +79,13 @@ usePointerEventHandler({ scene: scene.value, contextParts: context })
 
 const renderScene = () => {
     const container = scene.value as unknown as TresObject
-    render(h(Fragment, null, slots && slots.default ? slots.default() : []), container)
+    const internalFnComponent = defineComponent({
+        setup() {
+            provide('useTres', context);
+            return () => h(Fragment, null, slots && slots.default ? slots.default() : [])
+        }
+    })
+    render(h(internalFnComponent), container)
 }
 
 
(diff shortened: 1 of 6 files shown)


src/components/TresCanvas.vue
```diff
@@ -66,42 +65,31 @@ const slots = defineSlots<{
...
-    const context = useTresContextProvider({
-        scene: scene.value,
-        canvas: existingCanvas,
-        windowSize: props.windowSize,
-        disableRender,
-        rendererOptions: props,
-    })
 
-    usePointerEventHandler({ scene: scene.value, contextParts: context })
+onMounted(() => {
```
enpitsuLin, line 88: actually, I think all the staff in `onMounted` can be moved to the outside, cuz all side effects be processed in composable `useTresContextProvider`, only with caution

alvarosabu:
@Tinoooo what do you think?

enpitsuLin:
I found the scene in the playground package worked properly with `OrbitControls` locally, but in doc didn't, might have some other issues

src/composables/useCamera/index.ts
```diff
@@ -1,17 +1,18 @@
...
+export const useCamera = ({ sizes, scene }: Pick<TresContext, 'sizes' | 'scene'>) => {
+  const cameras = computed<Camera[]>(() => scene.value.children.filter((i): i is Camera => (i as any).isCamera))
   const camera = computed<Camera | undefined>(
-    () => cameras.value[0]
+    () => cameras.value[0],
   )
+  // don't known why need manually trigger here
+  watch(camera, () => {
+    triggerRef(cameras)
+  }, { deep: true })
```
enpitsuLin: just use `ShallowRef<Scene>` and it will track by manually `triggerRef`, so I refactor composables about using `scene`

***
but no idea about L13-15, if remove the camera should always be undefined, but this stuff doesn't make sense :P

src/components/TresCanvas.vue
```diff
@@ -131,17 +109,45 @@ onMounted(() => {
...
             'No camera found. Creating a default perspective camera. ' +
-            'To have full control over a camera, please add one to the scene.'
+            'To have full control over a camera, please add one to the scene.',
         )
         addDefaultCamera()
     }
-
-    if (import.meta.hot)
-        import.meta.hot.on('vite:afterUpdate', dispose)
 })
```
enpitsuLin, line 140: to void this warning is always shown, cuz this run before the camera is assigned. now check only run once by `triggerRef`

src/composables/useCamera/index.ts
alvarosabu: @enpitsuLin  scene is widely used in `cientos` , would the use of ShallowRef affect the reactivity of the abstractions?

alvarosabu:
Close in favor of https://github.com/Tresjs/tres/pull/365

PR #365, opened 2023-08-14
fix: 317 devtools

### LINKED ISSUE #317: Vue-devtools broken since v2 custom renderer
### Describe the bug

The official [Vue devtools](https://github.com/vuejs/devtools) is broken since v2. Most probably related to the edge case of the use of a custom renderer.

### Reproduction

https://stackblitz.com/edit/stackblitz-starters-sa3ruk?file=README.md

### Steps to reproduce

Open in a new tab
Check dev tools
Check error in console

```
TypeError: Cannot read properties of undefined (reading 'ownerDocument')
```

### System Info

_No response_

### Used Package Manager

pnpm

### Code of Conduct

alvarosabu:
Closes #317

### DIFF
--- a/src/composables/useCamera/index.ts
+++ b/src/composables/useCamera/index.ts
@@ -1,43 +1,39 @@
 import { Camera, OrthographicCamera, PerspectiveCamera } from 'three'
-import { computed, triggerRef, watch, watchEffect } from 'vue'
+import { computed, ref, triggerRef, watchEffect } from 'vue'
 
 import type { TresContext } from '../useTresContextProvider'
 
 
 export const useCamera = ({ sizes, scene }: Pick<TresContext, 'sizes' | 'scene'>) => {
   const cameras = computed<Camera[]>(() => scene.value.children.filter((i): i is Camera => (i as any).isCamera))
+  const activeCameraUuid = ref(cameras.value?.[0]?.uuid ?? "")
   const camera = computed<Camera | undefined>(
-    () => cameras.value[0],
+    () => cameras.value.find(c => c.uuid === activeCameraUuid.value) ?? cameras.value.at(0),
   )
-  // don't known why need manually trigger here
-  const unwatch = watch(camera, () => {
-    triggerRef(cameras)
-    unwatch?.()
-  }, { deep: true })
 
   const addCamera = (newCamera: Camera, active = false) => {
     if (cameras.value.some(({ uuid }) => uuid === newCamera.uuid))
       return
 
+    scene.value.add(newCamera)
+    triggerRef(scene)
     if (active)
       setCameraActive(newCamera)
-    else
-      cameras.value.push(newCamera)
   }
 
   const removeCamera = (camera: Camera) => {
     scene.value.remove(camera)
+    triggerRef(scene)
   }
 
   const setCameraActive = (cameraOrUuid: string | Camera) => {
-    const camera = cameraOrUuid instanceof Camera ?
-      cameraOrUuid :
-      cameras.value.find((camera: Camera) => camera.uuid === cameraOrUuid)
+    const cameraUuid = cameraOrUuid instanceof Camera ?
+      cameraOrUuid.uuid :
+      cameraOrUuid
 
-    if (!camera) return
+    if (!cameraUuid) return
 
-    const otherCameras = cameras.value.filter(({ uuid }) => uuid !== camera.uuid)
-    scene.value.add(...otherCameras)
+    activeCameraUuid.value = cameraUuid
   }
 
   watchEffect(() => {
--- a/src/composables/useCamera/index.ts
+++ b/src/composables/useCamera/index.ts
@@ -10,8 +10,9 @@ export const useCamera = ({ sizes, scene }: Pick<TresContext, 'sizes' | 'scene'>
     () => cameras.value[0],
   )
   // don't known why need manually trigger here
-  watch(camera, () => {
+  const unwatch = watch(camera, () => {
     triggerRef(cameras)
+    unwatch?.()
   }, { deep: true })
 
   const addCamera = (newCamera: Camera, active = false) => {
--- a/src/composables/useCamera/index.ts
+++ b/src/composables/useCamera/index.ts
@@ -1,17 +1,18 @@
-import { computed, watchEffect, onUnmounted, ref } from 'vue'
 import { Camera, OrthographicCamera, PerspectiveCamera } from 'three'
+import { computed, triggerRef, watch, watchEffect } from 'vue'
 
-import type { TresScene } from '../../types'
 import type { TresContext } from '../useTresContextProvider'
 
 
-export const useCamera = ({ sizes, scene }: Pick<TresContext, 'sizes'> & { scene: TresScene }) => {
-
-  // the computed does not trigger, when for example the camera postion changes
-  const cameras = ref<Camera[]>([])
+export const useCamera = ({ sizes, scene }: Pick<TresContext, 'sizes' | 'scene'>) => {
+  const cameras = computed<Camera[]>(() => scene.value.children.filter((i): i is Camera => (i as any).isCamera))
   const camera = computed<Camera | undefined>(
-    () => cameras.value[0]
+    () => cameras.value[0],
   )
+  // don't known why need manually trigger here
+  watch(camera, () => {
+    triggerRef(cameras)
+  }, { deep: true })
 
   const addCamera = (newCamera: Camera, active = false) => {
     if (cameras.value.some(({ uuid }) => uuid === newCamera.uuid))
@@ -21,11 +22,10 @@ export const useCamera = ({ sizes, scene }: Pick<TresContext, 'sizes'> & { scene
       setCameraActive(newCamera)
     else
       cameras.value.push(newCamera)
-
   }
 
   const removeCamera = (camera: Camera) => {
-    cameras.value = cameras.value.filter(({ uuid }) => uuid !== camera.uuid)
+    scene.value.remove(camera)
   }
 
   const setCameraActive = (cameraOrUuid: string | Camera) => {
@@ -36,7 +36,7 @@ export const useCamera = ({ sizes, scene }: Pick<TresContext, 'sizes'> & { scene
     if (!camera) return
 
     const otherCameras = cameras.value.filter(({ uuid }) => uuid !== camera.uuid)
-    cameras.value = [camera, ...otherCameras]
+    scene.value.add(...otherCameras)
   }
 
   watchEffect(() => {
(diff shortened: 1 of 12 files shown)


enpitsuLin:
@alvarosabu 

If using `useTresContextProvider` outside the `onMounted`, the WebGLRenderer constructor params will get an undefined **canvas** in composables `useRenderer`, so that will create an OffscreenCanvas as `renderer.domElem`, which be used to create `<OrbitControl/>` as args props, but after the canvas had been assigned the real canvas element, args prop of `<OrbitControl />` updated, then TresJS just assign props to the THREE.OrbitControl's args property in `patchProps` which wouldn't re-create it or update it (in [RendererOptions](https://github.com/Tresjs/tres/blob/211c3e172726a4cb339a02333ac596fcf3bb7793/src/core/nodeOps.ts#L212))

this refactor of the `useRenderer` will solve the renderer in `tresContext` always be defined that because it had initial with a `WebGLRenderer ` which renders on the wrong canvas

***

I found not just THREE.OrbitControl but also all THREE objects can't get proper behavior when the node gets args props to be updated, I think this is the fundamental reason.

alvarosabu:
> I found not just THREE.OrbitControl but also all THREE objects can't get proper behavior when the node gets args props to be updated, I think this is the fundamental reason. just check [this project](https://stackblitz.com/github/enpitsuLin/tres-issue-patch-props?file=src%2FBox.vue)

props that are arrays or vectors don't get updated reactively, is what you mean?

src/composables/useCamera/index.ts
```diff
@@ -1,42 +1,39 @@
...
   const camera = computed<Camera | undefined>(
-    () => cameras.value[0]
+    () => cameras.value.find(c => c.uuid === activeCameraUuid.value) ?? cameras.value.at(0),
   )
 
   const addCamera = (newCamera: Camera, active = false) => {
     if (cameras.value.some(({ uuid }) => uuid === newCamera.uuid))
       return
 
+    scene.value.add(newCamera)
```
Tinoooo, line 18: I think nothing in this composable should be responsible for adding things to the scene tree. That logic already exists here:
https://github.com/Tresjs/tres/blob/81b7914fc685aa7314541480a96801c0cc7fd542/src/core/nodeOps.ts#L111

Can you explain, why you added this code? Was something not working correctly?

src/composables/useRaycaster/index.ts
```diff
@@ -84,17 +84,20 @@ export const useRaycaster = (
...
-  canvas.value.addEventListener('pointermove', onPointerMove)
-  canvas.value.addEventListener('pointerleave', onPointerLeave)
-
-  onUnmounted(() => {
-    if (!canvas?.value) return
-    canvas.value.removeEventListener('pointerup', onPointerUp)
-    canvas.value.removeEventListener('pointerdown', onPointerDown)
-    canvas.value.removeEventListener('pointermove', onPointerMove)
-    canvas.value.removeEventListener('pointerleave', onPointerLeave)
+  watchPostEffect(onCleanUp => {
```
Tinoooo, line 87: Could you maybe explain to me, why this watcher is here now? Is there a need to add the eventlisteners again and again? 🤔

src/composables/useRenderer/index.ts
```diff
@@ -109,8 +109,8 @@ export function useRenderer(
     contextParts: { sizes, camera },
   }:
     {
-      canvas: MaybeRef<HTMLCanvasElement>
-      scene: Scene
+      canvas: MaybeRef<HTMLCanvasElement | undefined>
```
Tinoooo, line 112: Does this composable even make sense, when there is no canvas? 🤔

src/core/nodeOps.ts
```diff
@@ -188,9 +191,16 @@ export const nodeOps: RendererOptions<TresObject, TresObject> = {
         const prevNode = node as TresObject3D
         const prevArgs = _prevValue ?? []
         const args = nextValue ?? []
-
-        if (node.type && !deepArrayEqual(prevArgs, args)) {
-          root = Object.assign(prevNode, new catalogue.value[node.type](...nextValue))
+        const instanceName = node.userData.tres__name || node.type
```
Tinoooo, line 194: Can you maybe explain the new concept with `tres__name` to me? I don't grasp why this is here, yet.

Tinoooo (changes requested):

src/core/nodeOps.ts
alvarosabu: To re-instance THREE objects, we need to save their name ex `OrbitControls` on the `createElement` to be able to use it on the `patchProps` once args change.

This is only due to the fact that Three is not really consistent with properties across objects and Orbit controls doesn't have a `type` property 🤦🏻‍♂️

src/composables/useRaycaster/index.ts
enpitsuLin: this effect will just run once only the canvas is mounted and it is unmounted, that's similar to onMounted and onUnmounted, there are no one will trigger the canvas's effect,

src/composables/useCamera/index.ts
```diff
@@ -1,42 +1,39 @@
...
-
-  // the computed does not trigger, when for example the camera postion changes
-  const cameras = ref<Camera[]>([])
+export const useCamera = ({ sizes, scene }: Pick<TresContext, 'sizes' | 'scene'>) => {
+  const cameras = computed<Camera[]>(() => scene.value.children.filter((i): i is Camera => (i as any).isCamera))
+  const activeCameraUuid = ref(cameras.value?.[0]?.uuid ?? "")
   const camera = computed<Camera | undefined>(
-    () => cameras.value[0]
+    () => cameras.value.find(c => c.uuid === activeCameraUuid.value) ?? cameras.value.at(0),
   )
```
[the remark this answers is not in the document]
enpitsuLin: I thought this composable was used to manage all cameras in the scene, the state of the active camera, and operate cameras, eg. manually add a camera, or remove some cameras in the scene.

the previous logic created a state to manage that and do some hacky in userData (`tres__registerCamera` and `tres__deregisterCamera`) but it seems unreasonable and unnecessary but forgot to remove that, we just use the computed cameras and do `triggerRef` with the scene in certain necessary places

***
Actually, such things about those composables might be outbound of this contribution, but @alvarosabu said that's ok

src/composables/useRenderer/index.ts
```diff
@@ -109,8 +109,8 @@ export function useRenderer(
...
+      scene: ShallowRef<Scene>
```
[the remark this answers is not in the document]
enpitsuLin: yep

src/composables/useCamera/index.ts
Tinoooo: > I did some refactoring and realized the camera hasn't been truly added so I add this line
I don't think this is true. The camera ends up in the scene and I'm pretty sure it is because of this line:
https://github.com/Tresjs/tres/blob/81b7914fc685aa7314541480a96801c0cc7fd542/src/core/nodeOps.ts#L111
Could you maybe provide some code that proofs your statemen?

With the solution you proposed, cameras are added at the root level of the scene. So the component hierarchy is not respected anymore. The newly added code would prevent someone from adding a camera to a TresGroup for example.

src/core/nodeOps.ts
Tinoooo: Thank you for explaining @alvarosabu . I noticed, the reassignment happens even if there were no changes made in the args. I think this should not happen for performance reasons and because the disposal of some instances might be evaded. Wdyt?

src/composables/useCamera/index.ts
enpitsuLin: if add a camera by adding `<TresPerspectiveCamera/>` it goes here that's right, but if using `addCamera` which in this composable will never add to the scene(like addDefaultCamera), it works just because in the render use the camera to render the scene in `onLoop`.

simply, you can remove the code here

https://github.com/Tresjs/tres/blob/81b7914fc685aa7314541480a96801c0cc7fd542/src/composables/useRenderer/index.ts#L238

then create a scene without `<TresPerspectiveCamera/>` in TresCanvas, see we have no default camera has been added, just a warning and blank scene

> [TresJS ▲ ■ ●]  No camera found. Creating a default perspective camera. To have full control over a camera, please add one to the scene.

src/composables/useCamera/index.ts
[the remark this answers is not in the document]
Tinoooo: No, the composable was not used to add to the scene or remove cameras from the scene. As already stated [here](https://github.com/Tresjs/tres/pull/365#discussion_r1300254315) the responsibility lays in nodeOps.ts only.

`tres__registerCamera` and `tres__deregisterCamera` are not unreasonable and unnecessary. They exists so we can access the methods `addCamera` and `removeCamera` from the custom renderer. It is very similar to what happens in the `usePointerEventHandler. We haven't found a better solution than attaching those callbacks to properties of `userData` yet.

src/composables/useRaycaster/index.ts
Tinoooo: The code also runs, when making changes to the renderer (like toggling `stencil` for example), which is not necessary.

src/composables/useCamera/index.ts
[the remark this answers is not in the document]
enpitsuLin: I think the developer can use `addCamera` and `removeCamera` by using `useTres` that's right? but as I said it wouldn't truly add the camera to the scene

previous implements need to add `tres__registerCamera` and `tres__deregisterCamera` just to insure `camera` and `cameras` states will be set a correct value in TreContext

but after this, it just collects the child which was `Cameras` in the scene as a computed we needn't manually add cameras that were inserted by `nodeOps.ts` to the `cameras` state

src/composables/useRaycaster/index.ts
enpitsuLin: but the previous one has broken in some cases, maybe just using onMounted was ok

src/composables/useCamera/index.ts
enpitsuLin: yeah, the same camera would be added twice, that's my mistake, things about `tres__registerCamera` and `tres__deregisterCamera` should be removed after this

src/composables/useRaycaster/index.ts
alvarosabu: test it on `multiple` route

src/core/nodeOps.ts
alvarosabu: I will check it, 🤔 `deepArrayEqual` should avoid that, any specific instance or is happening to all of them?

enpitsuLin:
for left things from this PR, I will submit a clearer issue and a reproduction. (maybe tomorrow?🤔

Then we can discuss how to solve it.

PR #367, opened 2023-08-17
fix: added logic on pathProps to re-instance nodes once `args` change

### LINKED ISSUE #366: Re-instance THREE instances via args prop reactivity
### Description

As a developer using TresJS, I would like to change reactively the constructor params via args prop, which is not possible at the moment.

```ts
<script lang="ts" setup>
// renderer.domElement is undefined at the beginning. 
</script>

<template>
  <TresOrbitControls  :args="[activeCamera || camera, renderer?.domElement || domElement]" />
 </template>
 ```

### Suggested solution

Add logic to `nodeOps` to re-instance the THREE object when `args` change

### Alternative

_No response_

### Additional context

_No response_

### Validations

### PR DESCRIPTION
**Pull request by alvarosabu**
Closes #366

### DIFF
--- a/src/core/nodeOps.ts
+++ b/src/core/nodeOps.ts
@@ -15,6 +15,42 @@ function noop(fn: string): any {
   fn
 }
 
+function deepEqual(a, b) {
+  // If both are primitives, return true if they are equal
+  if (a === b) return true;
+
+  // If either of them is null or not an object, return false
+  if (a === null || typeof a !== "object" || b === null || typeof b !== "object") return false;
+
+  // Get the keys of both objects
+  const keysA = Object.keys(a), keysB = Object.keys(b);
+
+  // If they have different number of keys, they are not equal
+  if (keysA.length !== keysB.length) return false;
+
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/core/nodeOps.ts [hunk 1]
```diff
@@ -184,6 +184,14 @@ export const nodeOps: RendererOptions<TresObject, TresObject> = {
...
       let target = root?.[finalKey]
 
+      if (key === 'args' && !deepArrayEqual(_prevValue, nextValue)) {
+        const prevNode = node as TresObject3D
+
+        if (node.type && nextValue.length > 0) {
+          root = Object.assign(prevNode, new catalogue.value[node.type](...nextValue))
+        }
+      }
+
```
enpitsuLin, line 194: I think there should be premature return, or just add a condition on L220

enpitsuLin (changes requested):

src/core/nodeOps.ts
alvarosabu: Done

enpitsuLin (approved):

enpitsuLin (approved):

PR #369, opened 2023-08-18
fix: args should be empty when args was falsy

enpitsuLin:


src/core/nodeOps.ts
```diff
@@ -184,10 +184,12 @@ export const nodeOps: RendererOptions<TresObject, TresObject> = {
...
       let target = root?.[finalKey]
 
-      if (key === 'args' && !deepArrayEqual(_prevValue, nextValue)) {
+      if (key === 'args') {
         const prevNode = node as TresObject3D
+        const prevArgs = _prevValue ?? []
+        const args = nextValue ?? []
 
-        if (node.type && nextValue.length > 0) {
+        if (node.type && !deepArrayEqual(prevArgs, args)) {
```
alvarosabu, line 192: is  `!deepArrayEqual(prevArgs, args)` preventing the next line to happen if `nextValue` is an empty array? 

`new catalogue.value[node.type](...[])` will throw an error

alvarosabu (changes requested):

src/core/nodeOps.ts
enpitsuLin: @alvarosabu  I think it wouldn't throw any errors, just like L45, If it's not a problem there when `props.args` was an empty array, it's not a problem here.

alvarosabu (approved):

PR #390, opened 2023-09-03
chore: cleaned up playground and added leches GUI

alvarosabu:


playground/src/router.ts
```diff
@@ -6,20 +6,65 @@ const routes = [
...
+  {
+    path: '/conditional',
+    name: 'Conditional',
+    component: () => import('./pages/TheConditional.vue'),
+  },
   {
     path: '/multiple',
     name: 'Multiple',
-    component: () => import('./pages/multiple.vue'),
+    component: () => import('./pages/Multiple.vue'),
```
Tinoooo, line 32: Multiple.vue uses MultipleCanvas.vue, which doesn't exist anymore. MultipleCanvas was pretty useful in the past. It would be cool if we could keep it.

Tinoooo (changes requested):

playground/src/router.ts
alvarosabu: You are right, I deleted it by mistake, let me add it again

Tinoooo (approved):

PR #394, opened 2023-09-11
fix: proxy app context

alvarosabu:
This might solve https://github.com/Tresjs/nuxt/issues/33


alvarosabu:
Hi @enpitsuLin I really appreciate you taking the time, No worries. Get your stuff sorted out first. I will create a fake release on the next channel so we can test it in other repositories. 🤗

PR #401, opened 2023-09-21
fix(types): useTexture()

posva:
Correctly infer the return type based on the argument.

Note: I would personally have accepted a single string rather than `[string]` to build a single `Texture` but that would be a breaking change


alvarosabu:
@posva you're right, lets keep it like you changed for the moment. I will write it down for a potential v4 in the future.

alvarosabu:
Thanks for fixing this @posva

PR #404, opened 2023-09-22
feat: context (TresContext) is now exposed from TresCanvas

Tinoooo:


alvarosabu (approved):

alvarosabu:
@Tinoooo should we add this to the docs?

Tinoooo:
> @Tinoooo should we add this to the docs?

Yes! I forgot. I'm going to add the content now.

docs/api/composables.md
```diff
@@ -136,12 +136,12 @@ Then you can bind the textures to the material.
       <TresSphereGeometry />
       <TresMeshStandardMaterial
         :map="map"
-        :displacementMap="displacementMap"
-        :normalMap="normalMap"
-        :roughnessMap="roughnessMap"
-        :metalnessMap="metalnessMap"
-        :aoMap="aoMap"
-        :alphaMap="alphaMap"
+        :displacement-map="displacementMap"
```
JaimeTorrealba, line 139: This will break the property?

JaimeTorrealba (commented):
The addition of the tres-canvas page is a nice one!!

docs/api/composables.md
Tinoooo: I don't think so. It is not a nested property (https://threejs.org/docs/index.html#api/en/materials/MeshStandardMaterial.displacementMap).

docs/api/composables.md
```diff
@@ -179,19 +179,17 @@ watch(carRef, ({ model }) => {
 })
 ```
 
-## useTresContext (former useTres)
-
-This composable aims to provide access to the state model which contains the default renderer, camera, scene, and other useful properties.
+## useTresContext
+<!-- IMPROVE -->
```
JaimeTorrealba, line 183: We could remove this right?

docs/api/composables.md
```diff
@@ -179,19 +179,17 @@ watch(carRef, ({ model }) => {
...
+This composable aims to provide access to the state model which contains multiple useful properties.
 
 ```ts
-const { camera, renderer } = useTresContext()
+const { camera, renderer, camera, cameras } = useTresContext()
```
JaimeTorrealba, line 187: And here we have two cameras, the rest looks OK to me

docs/api/composables.md
```diff
@@ -209,3 +207,19 @@ import { useTresContext } from '@tresjs/core'
...
 ```
+
+### Properties of context
+| Property | Description |
+| --- | --- |
+| **camera** | the currently active camera |
+| **cameras** | the cameras that exist in the scene |
+| **controls** | the controls of your scene |
+| **deregisterCamera** | a method to deregister a camera. This is only required if you manually create a camera. Cameras in the template are deregistered automatically. |
+| **extend** | extends the component catalogue|
```
JaimeTorrealba, line 218: Here we can reference the extend docs https://tresjs.org/advanced/extending.html

JaimeTorrealba (approved):
I wrote some comments, but I'll approve, thanks for the docs, It's really useful for me :)

alvarosabu:
I like the new structure @Tinoooo very well done

docs/api/composables.md
Tinoooo: Yes. I removed it.

docs/api/composables.md
Tinoooo: camera is the active camera and camera**s** is the array of existing cameras.

docs/api/composables.md
Tinoooo: I added that.

docs/api/composables.md
JaimeTorrealba: Yeah but looks, it says, camera, render CAMERA, cameras, is that ok?

alvarosabu (approved):

PR #406, opened 2023-09-26
docs: fix invalid link in nav

kedniko:
Invalid link: https://tresjs.org/api/renderer.html


alvarosabu:
@kedniko thanks for the fix!

PR #428, opened 2023-10-29
fix(types): 427 vueprops change ref to vnoderef and key add nu…

### LINKED ISSUE #427: ref prop missing type for functional ref
### Describe the bug

Current type definitions don't allow for a function to be passed to the `ref` prop of a component, so any time you use a [functional ref](https://vuejs.org/guide/essentials/template-refs.html#function-refs) you'll run into a type error.

Apologies for not including a reproduction or sample but this should be quick to recreate in any dev environment with the reproduction steps.

### Reproduction

/

### Steps to reproduce

1. Install tresjs in any project
2. Add TresCanvas
3. Add new ref, and function to populate ref 
`const sphereRefs = ref<Object3D[]>([]);`
`const setSphereRefs = (el: any) => {
  if (el && !sphereRefs.value.includes(el)) {
    sphereRefs.value.push(el);
  }
};`
4. Add TresGroup, TresMesh, etc...
5. Pass a function to the `ref` property `:ref="(el) => "setSphereRef(el)"`
6. Type error pops up indicating this isn't a valid value for the `ref` prop despite this being a common use for the `ref` prop.

### System Info

_No response_

### Used Package Manager

yarn

### Code of Conduct

PatrickByrn:
types (fix #427)

Updated `ref` in VueProps to use `VNodeRef` type from vue, to support using a [function ref](https://vuejs.org/guide/essentials/template-refs.html#function-refs). Also updated key to accept numbers and symbols

https://github.com/Tresjs/tres/issues/427

### DIFF
--- a/src/types/index.ts
+++ b/src/types/index.ts
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/ban-types */
-import type { DefineComponent, Ref, VNode } from 'vue'
+import type { DefineComponent, VNode, VNodeRef } from 'vue'
 
 import type * as THREE from 'three'
 import type { EventProps as PointerEventHandlerEventProps } from '../composables/usePointerEventHandler'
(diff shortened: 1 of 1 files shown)


alvarosabu (approved):
@PatrickByrn thanks a lot of the fix 🙏🏻

PR #429, opened 2023-10-30
docs:add a translated user guide in Simplified Chinese

hawk86104:
To enable more Chinese users to use this library, a translated user guide in Simplified Chinese has been added. If there is time available in the future, we also plan to translate the API documentation.


alvarosabu:
Thanks for this @hawk86104 , we have a feature to enable docs in netlify to enable docs in different languages, I'm gonna prioritise it.

PR #432, opened 2023-11-01
refactor(TextureLoader): use THREE.DefaultLoadingManager as default loading manager

hawk86104:
I don't know if I should submit this type of pull request as a direct "create pull request" or as a "create draft pull request."


andretchen0:
@hawk86104 @alvarosabu 

Sorry for the delay. I'm not at home right now, so feedback will be slow.

I'm also not very clued into how Tres is working behind the scenes, so I'm not sure I'm the person to give a thumbs up or down here. @alvarosabu, I'm fine with whatever you decide.

----

Regarding the PR, I like the idea of using the THREE defaults. I wonder if we'd be better off in the long run if we used the [THREE `TextureLoader` signature](https://threejs.org/docs/?q=textureloader#api/en/loaders/TextureLoader). I.e., allow the user to pass their own `LoadingManager`. If they don't, use the default.

andretchen0:
FYI, updated the PR title [using the conventions her](https://www.conventionalcommits.org/en/v1.0.0/)e, in order to pass the linter check.

andretchen0 (commented):
I like the idea of using the THREE defaults. 

I wonder if we'd be better off in the long run if we used the [THREE `TextureLoader` signature](https://threejs.org/docs/?q=textureloader#api/en/loaders/TextureLoader). I.e., allow the user to pass their own `LoadingManager`. If they don't, use the default.

I'll go along with whatever @alvarosabu decides here.

hawk86104:
yeYes, I have submitted it ： https://github.com/Tresjs/tres/pull/471


| |
hawk
|
|
***@***.***
|


---- Replied Message ----
| From | Alvaro ***@***.***> |
| Date | 12/13/2023 16:52 |
| To | ***@***.***> |
| Cc | ***@***.***> ,
***@***.***> |
| Subject | Re: [Tresjs/tres] refactor(TextureLoader): use THREE.DefaultLoadingManager as default loading manager (PR #432) |

@andretchen0 I agree with your idea, I think I prefer the user to be able to pass the loading manager they want. @hawk86104 could you add that to the PR?

PR #433, opened 2023-11-03
feat(useSeek): Add `seekAll` and `seekAllByName` methods to `useSeek` composable

garrlker:
This implements [#236](https://github.com/Tresjs/cientos/issues/263)

Tldr; `seekAll` and `seekAllByName` are the same as `seek` and `seekByName` except that they return an array of all matches instead of the first match.

Both methods return an empty array `[]` when there are no matches

### DIFF
--- a/src/composables/useSeek/index.ts
+++ b/src/composables/useSeek/index.ts
@@ -9,6 +9,8 @@ import { useLogger } from '../useLogger'
 export interface UseSeekReturn {
   seek: (parent: THREE.Scene | THREE.Object3D, property: string, value: string) => THREE.Object3D | null
   seekByName: (parent: THREE.Scene | THREE.Object3D, value: string) => THREE.Object3D | null
+  seekAll: (parent: THREE.Scene | THREE.Object3D, property: string, value: string) => THREE.Object3D[]
+  seekAllByName: (parent: THREE.Scene | THREE.Object3D, value: string) => THREE.Object3D[]
 }
 
 /**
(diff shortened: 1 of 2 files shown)


andretchen0:
I updated the PR title, so the linter doesn't flag it.

Just FYI, [here's the expected format for titles.](https://www.conventionalcommits.org/en/v1.0.0/)

src/composables/useSeek/index.ts
```diff
@@ -44,6 +46,30 @@ export function useSeek(): UseSeekReturn {
...
+ * @param {(THREE.Scene | THREE.Object3D)} parent
+ * @param {string} property
+ * @param {string} value
+ * @return {*}  {(THREE.Object3D[])}
+ */
+  function seekAll(parent: THREE.Scene | THREE.Object3D, property: string, value: string): THREE.Object3D[] {
+    const foundChildren: THREE.Object3D[] = []
+
+    parent.traverse((child) => {
+      if ((child as any)[property] === value) {
```
alvarosabu, line 61: Since we are looking by search term 'Cloth' for example, if we have objects with names 'Cloth001', 'Cloth002' this evaluation would not grab them.

You can use `includes` for this
```
if ((child as any)[property].includes(value)) {`
```

alvarosabu (changes requested):
We are almost there @garrlker:
- Please add the documentation for this composable on `docs/guide/use-seek.md` 
- Check the comment on the evaluation of strings

garrlker:
@andretchen0 Thanks! I'll keep that in mind. Also I've never seen the conventional commits site before, double thanks for that!

@alvarosabu I updated the documentation with a very barebones example/description and used `includes` over `===`

alvarosabu (approved):

PR #439, opened 2023-11-10
fix: TresGroup components emit children mesh events

garrlker:
Not quite done yet. It works, but only for children components that also have event handlers wired onto them. 
It's seems to be ignoring the rest of the children.

Also random but it seems like mouse events are firing off sporadically?

I'm including TheExperience bug recreation from the issue for now, but will remove it before merge


alvarosabu:
Hey buddy @garrlker any update on this one?

Tinoooo:
@garrlker I finally found some time to check out your PR. Sorry for the huge delay.

I have two topcs I want to discuss.

1️⃣ Parent elements (the group) only receive events if the child element has event listeners attached to them. I know that Threlte and R3F work this way, but I think this behavior is not quite intuitive to users. But this might be just me. 

2️⃣ Using `pointer-leave` on a TresGroup also triggers, when the mouse leaves a child object but not the group. Users might find this behavior unintuitive.

Both topics require us to first define how the event handling/bubbling should work. What are you opinions about this @garrlker @alvarosabu @JaimeTorrealba? If this discussion requires extensive time, we could schedule a call to address it.

alvarosabu:
Hi, @Tinoooo thanks for jumping in, and thanks both @garrlker for taking care of this one.

I have done benchmarking and created a simple app on the 3 major renderers to check the differences between them and understand the problem (without taking this PR into consideration):

R3F and Threlte share the most similarities, especially on the event topic vs Tres. This is not necessarily a bad thing but I think we should agree and maybe ask the community what behavior they are expecting to decide if its a bug.

With that being said I think its an area with a lot of room for improvement. I'm up for a pair programming session together if you like! We can do it on discord on the after-work channel or even a bug-hunting stream on Twitch, as you prefer.

Let me know what do you think

alvarosabu:
@garrlker quick small feedback:

1.  when creating a branch please create it under `feature` or `bugfix` depending on the nature of the issue 😊, this branch should be `bugfix/426-groups-emit...` 
2. I normally recommend to create a route on the playground to test out instead of using the `TheExperience`, specially useful for code reviews and avoid conflicts.

garrlker:
@Tinoooo thank you for diving into this so thoroughly. I missed those cases

100% agree on us clearly defining how this should work and asking the community.
IMO while we don't *have* to follow R3F/Threlte if they are both aligning on similar patterns then we should follow common convention.

We can still implement it in a more Vue-ish way, Ex. we should have `.prevent` and `.stop` event modifiers to handle Threejs boilerplate. There are possibly other common patterns we could shorten, but those 2 are easy layups

@alvarosabu Gotcha, ty for the tips

I'm also down for a call and/or pair programming to get this finished

garrlker:
Closing for #515 

Will continue work there

PR #456, opened 2023-11-30
fix: incorrect MathRepresentation type

alexzhang1030:
Due to the https://github.com/three-types/three-ts-types/pull/615 PR resolving the incorrect type of `Color`, we cannot use `set(...args: number[]): any` to represent the `Color` class.


alvarosabu (approved):
Thanks so much for this fix @alexzhang1030

PR #466, opened 2023-12-10
fix: 449 memory management

### LINKED ISSUE #449: Memory leak caused by non disposed scene, when using routes
### Describe the bug

In our project, we use a bit big bigger models and we test our app on mobile browsers. While observing the memory, it became clear that unmounting the TresCanvas, does not properly dispose the scene and some (maybe all) internal objects. So navigating in and out a few times on iOS device, cause the page to crash and reload, because it runs out of memory.

Screenshot 2023-11-19 at 20 15 16

Unfortunately, my understanding of three.js is not much, but I had similar issues in the past and it was related to the way the scene and objects, that were added to the scene, being not properly disposed. Its possible that some references to some internal TresJS objects are also the cause of the issue. I just dont know.

Its also possible that I am missing something in the way I use the framework (which will be the best for me), so hopefully someone can give me a hand with this.

Thank you!

### Reproduction

https://stackblitz.com/edit/tresjs-basic-tmk778

### Steps to reproduce

To reproduce, open the reproduction link in a Chromium browser and use the `Memory` tab in the dev tools. 
At the top of the web page, you will find a basic navigation, that routes into two separate pages, where "The Experience" loads the TresCanvas and "Home" is used to unmount the other one.

### System Info

_No response_

### Used Package Manager

npm

### Code of Conduct

alvarosabu:
Potentially closes #449 

- Recursive removal of nodes dispose (GPU) of all three disposable properties and removes refs (CPU) **if the node is not primitive**
- If it is a primitive, the user needs to handle the disposal manually by using a `dispose` function

### DIFF
--- a/src/core/nodeOps.ts
+++ b/src/core/nodeOps.ts
@@ -115,71 +115,90 @@ export const nodeOps: RendererOptions<TresObject, TresObject> = {
   },
   remove(node) {
     if (!node) return
-    // remove is only called on the node being removed and not on child nodes.
 
-    if (node.isObject3D) {
-      const object3D = node as unknown as Object3D
+    const disposeMaterialsAndGeometries = (object3D: Object3D) => {
+      const tresObject3D = object3D as TresObject3D
 
-      const disposeMaterialsAndGeometries = (object3D: Object3D) => {
-        const tresObject3D = object3D as TresObject3D
-
-        if (!object3D.userData.tres__materialViaProp) {
-          tresObject3D.material?.dispose()
-          tresObject3D.material = undefined
-        }
-          
-        if (!object3D.userData.tres__geometryViaProp) {
-          tresObject3D.geometry?.dispose()
-          tresObject3D.geometry = undefined
-        }
+      if (!object3D.userData.tres__materialViaProp) {
+        tresObject3D.material?.dispose()
+        delete tresObject3D.material
       }
+        
+      if (!object3D.userData.tres__geometryViaProp) {
+        tresObject3D.geometry?.dispose()
+        delete tresObject3D.geometry
+      }
+    }
 
-      const deregisterAtPointerEventHandler = scene?.userData.tres__deregisterAtPointerEventHandler
-      const deregisterBlockingObjectAtPointerEventHandler
+    const deregisterAtPointerEventHandler = scene?.userData.tres__deregisterAtPointerEventHandler
+    const deregisterBlockingObjectAtPointerEventHandler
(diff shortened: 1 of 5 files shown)


src/core/nodeOps.ts
```diff
@@ -115,71 +115,90 @@ export const nodeOps: RendererOptions<TresObject, TresObject> = {
...
-      disposeMaterialsAndGeometries(object3D)
-      deregisterCameraIfRequired(object3D)
-      deregisterAtPointerEventHandlerIfRequired?.(object3D as TresObject)
     }
 
-    node.dispose?.()
+    // Only dispose if it's not a primitive and it has a dispose method,
+    //  if its a primitive, disposal needs to be handled by the user.
+    if (!isPrimitive && node?.isObject3D) {
+      node.children.forEach((child: TresObject) => nodeOps.remove(child))
```
Tinoooo, line 185: Shouldn't this always happen? Even if the node is a primitive or an Object3D the children should be disposed properly right?

src/core/nodeOps.ts
```diff
@@ -115,71 +115,90 @@ export const nodeOps: RendererOptions<TresObject, TresObject> = {
...
+          delete node[p]
+        }
+      }
+    }
+    
+    if (shouldDispose && node.dispose && node.type !== 'Scene') {
+      node.dispose()
+    }
+
+    // Add invalidation code similar to https://github.com/pmndrs/react-three-fiber/blob/master/packages/fiber/src/core/utils.ts#L436
```
Tinoooo, line 201: Is this an open todo?

alvarosabu:
@Tinoooo just to give you an update I'm waiting to merge #514 to v4 before continuing with this one.

src/core/nodeOps.ts
andretchen0: @Tinoooo 

I think the idea is that a primitive's `:object` is created by the user in `<script setup>`. If Tres cleans it up, it risks breaking the user's setup. 

E.g., the user might have plans to reuse the `:object` after the `<primitive />` has been removed.

[Fwiw, R3F seems to follow the same rules.
](https://github.com/pmndrs/react-three-fiber/blob/0ef66a1d23bf16ecd457dde92b0517ceec9861c5/packages/fiber/src/core/renderer.ts#L207)

alvarosabu: Yep basically what @andretchen0 stated.

Tinoooo: Got ya 👍

alvarosabu:
Closed in favor of #606

PR #467, opened 2023-12-10
fix: click event is called without click

Tinoooo:


alvarosabu (approved):

alvarosabu:
@Tinoooo thanks for the fix 💚

PR #479, opened 2023-12-21
feat: 474 vue chrome devtools plugin

alvarosabu:


src/devtools/plugin.ts
```diff
@@ -0,0 +1,295 @@
...
+      
+      api.on.getInspectorState((payload) => {
+        if (payload.inspectorId === INSPECTOR_ID) {
+          // Your logic here
+          const [instance] = tres.scene.value.getObjectsByProperty('uuid', payload.nodeId)
+          if (!instance) return 
+          if (prevInstance && highlightMesh && highlightMesh.parent) {
+            prevInstance.remove(highlightMesh)
+          }
+          const newHighlightMesh = createHighlightMesh(instance)
```
enpitsuLin, line 185: for highlight mesh, maybe we can mark it, then filter in `createNode`?

PS: I think maybe we can ship all animation logic to a new Mesh class extends from `THREE.Mesh`, then this new Mesh class naturally becomes a marker

Issue #489, opened 2024-01-02
Vue devtools 


alvarosabu:
### Description

As a developer, I would like to have the same DX devtools experience provided by the [Nuxt devtools ](https://devtools.nuxt.com/) for vue using https://github.com/vuejs/devtools-next

### Suggested solution

Custom tab at https://github.com/vuejs/devtools-next

### Validations

- [X] I agree to follow this project's [Code of Conduct](https://github.com/Tresjs/tres/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/Tresjs/tres/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://tresjs.org/guide).
- [X] Check that there isn't [already an issue](https://github.com/tresjs/tres/issues) that reports the same bug to avoid creating a duplicate.

Issue #487, opened 2024-01-02
Types issues after v156


alvarosabu:
### Description

`r156` added some type changes which break our types. The generic type Object3D expects types which extent Object3DEventMap

Check this along to new potential breaking changes 

### Suggested solution

Fix types

### Validations

- [X] I agree to follow this project's [Code of Conduct](https://github.com/Tresjs/tres/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/Tresjs/tres/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://tresjs.org/guide).
- [X] Check that there isn't [already an issue](https://github.com/tresjs/tres/issues) that reports the same bug to avoid creating a duplicate.

Issue #474, opened 2023-12-18
Vue chrome devtools plugin


alvarosabu:
### Description

It's time to create a plugin for the official vue devtools 

### Suggested solution

https://devtools.vuejs.org/plugin/plugins-guide.html

### Validations

- [X] I agree to follow this project's [Code of Conduct](https://github.com/Tresjs/tres/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/Tresjs/tres/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://tresjs.org/guide).
- [X] Check that there isn't [already an issue](https://github.com/tresjs/tres/issues) that reports the same bug to avoid creating a duplicate.

Issue #426, opened 2023-10-27
TresGroups Pointer events missing


alvarosabu:
### Describe the bug

Pointer events on `<TresGroup>`  are not working, clicking on the children should trigger a click event

### Reproduction

https://stackblitz.com/edit/tresjs-basic-bivxri?file=README.md,src%2Fcomponents%2FTheExperience.vue

### Steps to reproduce

Add a pointer event to a  `<TresGroup>` 

### System Info

```shell
System:
    OS: Linux 5.0 undefined
    CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 18.18.0 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 9.4.2 - /usr/local/bin/npm
    pnpm: 8.9.2 - /usr/local/bin/pnpm
  npmPackages:
    @tresjs/cientos: ^3.5.1 => 3.5.1 
    @tresjs/core: ^3.4.1 => 3.4.1 
    @tresjs/eslint-config-vue: ^0.2.1 => 0.2.1 
    vite: ^4.5.0 => 4.5.0 
```
```


### Used Package Manager

pnpm

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/Tresjs/tres/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/Tresjs/tres/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://tresjs.org/guide).
- [X] Check that there isn't [already an issue](https://github.com/tresjs/tres/issues) that reports the same bug to avoid creating a duplicate.
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

CptJJ:
Is there any progress on this?

garrlker:
@CptJJ hey, I've got a draft PR up [here](https://github.com/Tresjs/tres/pull/439)

@Tinoooo is going to review it when he has time. Feel free to pull and test my branch locally by running the playground

garrlker:
The scope of this issue grew quite a bit and this will be handled in a another PR I'm working on for #515

PR #490, opened 2024-01-02
feat: release v4

alvarosabu:
Includes: 

- [x] Three v160 and Vue v3.4 #488
- [x] #140 
- [x] Vue chrome devtools plugin #474  #479 
- [x] #426 #439 
- [x] #402 
- [x] #487 
- [x] ~~#489~~
- [x] #466


andretchen0:
This is a really beefy PR. SInce it's still a draft, is there anything we should concentrate on at the moment?

alvarosabu:
@andretchen0 actually, all the things that have been merged here already passed their own PRs, so this is only for organizing, no need to extra check it.

Something that we would really need support with is #515, I already aligned with @garrlker yesterday, and he will take a refresh this weekend to see what's left, could you support us there? This is the last feature we need for v4.

PR #498, opened 2024-01-07
feat: 492 set tone mapping default to acesfilmictonemapping

### LINKED ISSUE #492: Set tone-mapping default to ACESFilmicToneMapping
### Describe 

As a developer using TresJS, I would like the tone-mapping of the renderer to be set to `ACESFilmicToneMapping` as default.

This will introduce a Breaking change.

Decisions based on feedback from the community: 

Screenshot 2024-01-04 at 08 32 46

Screenshot 2024-01-04 at 08 33 36

From left to right: r3f, Tres, Threlte

Screenshot 2024-01-03 at 15 56 42

## Docs

[Render defaults](https://docs.pmnd.rs/react-three-fiber/api/canvas#render-defaults)
Canvas uses [createRoot](https://docs.pmnd.rs/react-three-fiber/api/canvas#createroot) which will create a translucent THREE.WebGLRenderer with the following constructor args:

antialias=true
alpha=true
powerPreference="high-performance"
and with the following properties:

outputColorSpace = THREE.SRGBColorSpace
toneMapping = THREE.ACESFilmicToneMapping
It will also create the following scene internals:

A THREE.Perspective camera
A THREE.Orthographic cam if orthographic is true
A THREE.PCFSoftShadowMap if shadows is true
A THREE.Scene (into which all the JSX is rendered) and a THREE.Raycaster
In recent versions of threejs, THREE.ColorManagement.enabled will be set to true to enable automatic conversion of colors according to the renderer's configured color space. R3F will handle texture color space conversion. For more on this topic, see https://threejs.org/docs/#manual/en/introduction/Color-management.

alvarosabu:
Closes #492

### DIFF
--- a/src/composables/useRenderer/const.ts
+++ b/src/composables/useRenderer/const.ts
@@ -1,4 +1,4 @@
-import { ACESFilmicToneMapping, PCFSoftShadowMap, SRGBColorSpace } from 'three'
+import { ACESFilmicToneMapping, NoToneMapping, PCFSoftShadowMap, SRGBColorSpace } from 'three'
 
 export const rendererPresets = {
   realistic: {
@@ -11,6 +11,10 @@ export const rendererPresets = {
       enabled: true,
       type: PCFSoftShadowMap,
     },
+  },
+  flat: {
+    toneMapping: NoToneMapping,
+    toneMappingExposure: 1,
   },
 }
 
--- a/src/composables/useRenderer/index.ts
+++ b/src/composables/useRenderer/index.ts
@@ -1,4 +1,4 @@
-import { Color, WebGLRenderer } from 'three'
+import { ACESFilmicToneMapping, Color, WebGLRenderer } from 'three'
 import { shallowRef, watchEffect, onUnmounted, type MaybeRef, computed, watch } from 'vue'
 import {
   toValue,
@@ -122,7 +122,9 @@ export function useRenderer(
 ) {
 
   const webGLRendererConstructorParameters = computed<WebGLRendererParameters>(() => ({
-    alpha: toValue(options.alpha),
+    alpha: toValue(options.alpha) === undefined // an opinionated default of tres
+      ? true
+      : toValue(options.alpha),
     depth: toValue(options.depth),
     canvas: unrefElement(canvas),
     context: toValue(options.context),
@@ -179,6 +181,8 @@ export function useRenderer(
 
   const threeDefaults = getThreeRendererDefaults()
 
+  // Set opinionated defaults
+  
   watchEffect(() => {
     const rendererPreset = toValue(options.preset)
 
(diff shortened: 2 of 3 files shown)


src/composables/useRenderer/const.ts
```diff
@@ -12,6 +12,10 @@ export const rendererPresets = {
       type: PCFSoftShadowMap,
     },
   },
+  flat: {
+    toneMapping: NoToneMapping,
```
alvarosabu, line 16: I added it here, but is not working at all when you set the preset to `flat` 

@Tinoooo should we tackle it here or in a separate PR?

src/composables/useRenderer/index.ts
```diff
@@ -122,7 +122,9 @@ export function useRenderer(
 ) {
 
   const webGLRendererConstructorParameters = computed<WebGLRendererParameters>(() => ({
-    alpha: toValue(options.alpha),
+    alpha: toValue(options.alpha) === undefined // an opinionated default of tres
```
andretchen0, line 125: `toValue` returns `undefined` or a `boolean`, so it would be shorter and maybe more idiomatic to use a [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) instead of a ternary:

```ts
alpha = toValue(options.alpha) ?? true, // an opinionated default of tres
```

src/composables/useRenderer/index.ts
```diff
@@ -214,7 +218,7 @@ export function useRenderer(
       set(renderer.value, pathInThree, getValue(option, pathInThree))
 
     setValueOrDefault(options.shadows, 'shadowMap.enabled')
-    setValueOrDefault(options.toneMapping, 'toneMapping')
+    setValueOrDefault(options.toneMapping || ACESFilmicToneMapping, 'toneMapping')
```
andretchen0, line 221: Is completely disallowing `NoToneMapping` the desired behavior here?

If not, it seems to me that this should use a nullish coalescing operator instead. E.g.,

```ts
setValueOrDefault(options.toneMapping ?? ACESFilmicToneMapping, 'toneMapping')
```

### Rationale

Logical OR (`||`) [returns the right side if the left side is falsy.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_OR#description) `0` is falsy.

Three's [`NoToneMapping` constant has a value of `0`](https://github.com/mrdoob/three.js/blob/d04539a76736ff500cae883d6a38b3dd8643c548/src/constants.js#L54), i.e., falsy.

So ...

```ts
NoToneMapping || ACESFilmicToneMapping === ACESFilmicToneMapping // true
```

src/composables/useRenderer/index.ts
```diff
@@ -1,4 +1,4 @@
-import { Color, WebGLRenderer } from 'three'
+import { ACESFilmicToneMapping, Color, WebGLRenderer } from 'three'
```
andretchen0, line 1: (I can't add a comment there because it wasn't changed in this PR, but ...)

... It looks like the comment on line 80 is now out of sync:

```jsdoc
   * @default NoToneMapping
```

I think it should now be:

```jsdoc
   * @default ACESFilmicToneMapping
```

andretchen0 (changes requested):
A couple of points to look over before merging.

alvarosabu:
Hey @andretchen0 thanks for checking the PR and for the feedback, I just pushed the changes you requested. Let me know if all good

andretchen0 (approved):
Sorry for the delay. I didn't see the review request notification.

Looks good!

PR #500, opened 2024-01-07
chore: minor code improvements

alvarosabu:


src/composables/useLogger.ts
```diff
@@ -4,30 +4,30 @@ export const isProd = import.meta.env.MODE === 'production'
...
   }
   /*eslint-enable no-console */
 
   return {
-    logError,
-    logWarning,
-    logMessage,
+    err,
+    warn,
+    msg,
```
Tinoooo, line 31: I liked the previous names a little better because it did represent a bit better what happens, but I am also fine with these.

Tinoooo (approved):

garrlker (approved):

src/composables/useLogger.ts
```diff
@@ -4,30 +4,30 @@ export const isProd = import.meta.env.MODE === 'production'
...
 
-  function logMessage(name: string, value: any) {
+  function msg(name: string, value: any) {
     if (!isProd) {
       console.log(`${logPrefix} - ${name}:`, value)
     }
   }
   /*eslint-enable no-console */
 
   return {
```
andretchen0, line 28: `err` and `msg` are nouns. It's clearer to me when functions/methods are (or start with) verbs, like `logError`, `logMessage`.

Also, this changes the API, so we'll have to change the user code in Cientos as well. 

I won't block the PR, but I prefer the status quo here.

alvarosabu: Fair enough @andretchen0 @Tinoooo  I will close this one then

andretchen0:
@alvarosabu 

Just thinking out loud here.

I wonder if doing this in JS/TS is the right approach. The goal of `useLogger` as I understand it is that we want to add a Tres prefix to `console.error`, `console.warn`, and `console.log`. The other functionality, like narrowing the accepted arguments, is maybe less important.

From the perspective of a developer, I'd prefer just to use `console.[whatever]`. That's what's ingrained in my brain and that's what I usually type, before eventually erasing and using `useLogger().[whatever]`.

That would have the advantage over a JS/TS solution in that:

* For `useLogger().LogError`, the JS/TS solution adds a first line to the stack trace, pointing at `useLogger`. The caller – where the actual error was thrown – is pushed to the second line.
* For `useLogger().[LogError, LogWarning, LogMessage]`, the JS/TS solution gives the filename/line number of `useLogger` instead of the filename/line number of the caller.

Using `console.[whatever]` would avoid those inconveniences for the end user.

So, maybe we find/write a Vite plugin to insert the prefix into `console.[whatever](PREFIX, ...args)` when the code is bundled?

andretchen0:
@alvarosabu 

[Following up from this comment above.](https://github.com/Tresjs/tres/pull/500#issuecomment-1902668833)

I have a WIP for a Rollup plugin to add a first argument to `console.[log | info | warn | error]` at bundle time. It's a standalone plugin that could be included in `vite.config.ts` of the various Tres projects. 

For example:

```js
const s = "my string"
export default function () {
  console.log(2, s);
}

[1,2,3].forEach(n => console.info(n))

try {
    JSON.parse('{}')
} catch(e) {
    console.error(e)
}
```

becomes

```js
const s = "my string";
function main () {
  console.log('[TresJS ▲ ■ ●]', 2, s);
}

[1,2,3].forEach(n => console.info('[TresJS ▲ ■ ●]', n));

try {
    JSON.parse('{}');
} catch(e) {
    console.error('[TresJS ▲ ■ ●]', e);
}
```

The prefix can be set in `package.json`, so Cientos could do ...

```js
console.error(e)
```

```js
console.error('[TresJS/Cientos ▲ ■ ♥]', e)
```

## Questions

* Is it worthwhile to continue exploring this?
* Is it worthwhile putting it in a separate `Tres / *` repo? And NPM package? That way it could be imported in our `package.json`s.

alvarosabu:
> I have a WIP for a Rollup plugin to add a first argument to console.[log | info | warn | error] at bundle time. It's a standalone plugin that could be included in vite.config.ts of the various Tres projects.

Actually, I think this could be a really useful plugin for OSS authors outside of TresJS 🤔, what if you launch an `unplugin`?

While looking for the unplugin github repo I found this gem https://github.com/unplugin/unplugin-turbo-console wdyt?

andretchen0:
@alvarosabu 

No luck with TurboConsole. As far as I can see, [it's only designed to work with `console.log`](https://github.com/unplugin/unplugin-turbo-console/blob/085af5f73007e092750d6e558563bc4d75b97e4e/src/core/transform/common.ts#L61). So we can't mimic all the functionality in `useLogger`.

Preferences on how to proceed?

Issue #511, opened 2024-01-14
The screen flickers severely when the resize event is triggered


Steve245270533:
### Describe the bug

When the resize event is triggered, the canvas is not re-rendered in time, causing the screen to flicker. This situation is also seen in the official example.

![Screen flicker1](https://github.com/Tresjs/tres/assets/46118247/6f068e3f-a00f-4e5c-a43e-0faa1afebcaf)  

![Screen flicker2](https://github.com/Tresjs/tres/assets/46118247/8ba48e26-aa08-449c-8d04-74965cc8267c)

### Reproduction

https://stackblitz.com/edit/stackblitz-starters-hhwpaq?file=src%2Fcomponents%2FTheExperience.vue

### Steps to reproduce

This happens even when building the most basic scene：
```vue
<template>
	<TresCanvas ref="canvasRef" v-bind="gl">
		<TresAxesHelper />
	</TresCanvas>
</template>

<script setup lang="ts">
	import { BasicShadowMap, SRGBColorSpace, NoToneMapping } from 'three';

	const gl = {
		clearColor: '#82DBC5',
		shadows: true,
		alpha: false,
		shadowMapType: BasicShadowMap,
		outputColorSpace: SRGBColorSpace,
		toneMapping: NoToneMapping,
	};

</script>
```

### System Info

```shell
System:
    OS: Windows 10 10.0.19045
    CPU: (8) x64 Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
    Memory: 2.49 GB / 15.90 GB
  Binaries:
    Node: 18.19.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.17 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 9.6.6 - ~\AppData\Roaming\npm\npm.CMD
    pnpm: 7.7.0 - ~\AppData\Roaming\npm\pnpm.CMD
  Browsers:
    Edge: Chromium (120.0.2210.133)
    Internet Explorer: 11.0.19041.3636
  npmPackages:
    @tresjs/core: ^3.6.0 => 3.6.0
```

### Used Package Manager

pnpm

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/Tresjs/tres/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/Tresjs/tres/blob/main/CONTRIBUTING.md).
[...]

PR #512, opened 2024-01-16
fix(usetrescontextprovider): fixed rendering issues caused when resize is triggered

Steve245270533:
Solved this issue (#511)

By minimizing the scene reproduction, I found that the real cause of jitter during rendering is the use of the two APIs `useElementSize` / `useWindowSize`. 

They both use the [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) API, which is always a few milliseconds slower than the triggering of the onResize event. 

I also compared the behavior of R3F when resizing and found that there is an anti shake operation. 

Therefore, we only need to perform an anti-shake operation on the width and height values ​​to solve the problem.

### DIFF
--- a/src/composables/useTresContextProvider/index.ts
+++ b/src/composables/useTresContextProvider/index.ts
@@ -59,7 +59,7 @@ export function useTresContextProvider({
     width: 0,
     height: 0,
   })
-  const DebouncedReactiveSize = refDebounced(reactiveSize, 10)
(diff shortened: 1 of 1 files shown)


alvarosabu:
Hello @Steve245270533 thank you very much for taking the time to open a ticket and create a PR with the fix, looks pretty solid, the debounce makes sense and it will also improve the performance 

I will test it and give you feedback 🙏🏻

alvarosabu (approved):

PR #513, opened 2024-01-17
refactor: Simplify the logic for calculating sizes and extract it into useSizes

Steve245270533:
Attempting to simplify the logic related to `sizes` and extract it into `useSizes` for easier maintenance in the future.

### DIFF
--- a/src/composables/useSizes/index.ts
+++ b/src/composables/useSizes/index.ts
@@ -11,14 +11,14 @@ export interface SizesType {
 export default function useSizes(
   windowSize: MaybeRefOrGetter<boolean>,
   canvas: MaybeRef<HTMLCanvasElement>,
-  debounce: number = 10,
+  debounceMs: number = 10,
 ) {
   const reactiveSize = toValue(windowSize)
     ? useWindowSize()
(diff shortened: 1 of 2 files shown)


Tinoooo (approved):

garrlker (approved):

src/composables/useSizes/index.ts
```diff
@@ -0,0 +1,30 @@
...
+export interface SizesType {
+  height: Readonly<Ref<number>>
+  width: Readonly<Ref<number>>
+  aspectRatio: ComputedRef<number>
+}
+
+export default function useSizes(
+  windowSize: MaybeRefOrGetter<boolean>,
+  canvas: MaybeRef<HTMLCanvasElement>,
+  debounce: number = 10,
```
andretchen0, line 14: A naming suggestion. 

Background: Vueuse calls this argument `ms`. 

Maybe we call it `debounceMs`? That makes it clear that it's expecting millisecond units, and not e.g., fractions of a second or a boolean.

Steve245270533: Oh, I see, that makes sense. I should indeed give them more accurate names. I'll go make the changes.

PR #514, opened 2024-01-18
feat: 503 conditional rendering of primitives

### LINKED ISSUE #503: Conditional and dynamic rendering of primitives.
### Describe the bug

Reported by Guillaume Leo on discord:

- Object prop is not reactive
- Conditional rendering is not working. `v-if` 

https://discord.com/channels/1047126995424780288/1193882152811642892

### Reproduction

https://stackblitz.com/fork/github/stackblitz/starters/tree/main/tres?title=TresJS&description=Quick%20start%20repo%20for%20Tres.js%20projects

### Steps to reproduce

_No response_

### System Info

_No response_

### Used Package Manager

npm

### Code of Conduct

alvarosabu:
Closes #503 

- [x] `Object` prop is not reactive
- [x] Conditional rendering is not working.` v-if`


andretchen0 (approved):
Glad to have this update!

I only found a small bit of code that looks like it can be simplified. I won't hold up the PR.

andretchen0:
@alvarosabu 

Do you need a hand with the failing tests and linter?

src/core/nodeOps.ts
```diff
@@ -39,7 +39,8 @@ export const nodeOps: RendererOptions<TresObject, TresObject> = {
       if (props?.object === undefined) logError('Tres primitives need a prop \'object\'')
       const object = props.object as TresObject
```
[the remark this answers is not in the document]
alvarosabu: Hey @andretchen0, thanks for checking it, I ended up modifying this part because we need to clone the `prop.object`  to avoid overwriting the original three object when swapping happens.

alvarosabu:
It's pretty much done @andretchen0 if you take another look before merging 🙏🏻

src/core/nodeOps.ts
```diff
@@ -39,7 +39,8 @@ export const nodeOps: RendererOptions<TresObject, TresObject> = {
...
       name = object.type
-      instance = Object.assign(object, { type: name, attach: props.attach, primitive: true })
+      instance = Object.assign(object.clone(), { type: name }) as TresObject
```
andretchen0, line 42: I'm not sure if there's a workaround, but using a clone here means that references in the component's setup won't work.

I've added a comment to playground primitives.vue file with some code that demonstrates the problem.

playground/src/pages/primitives.vue
```diff
@@ -0,0 +1,103 @@
...
+  }),
+)
+
+const sphere = new Mesh(
+  new SphereGeometry(1, 32, 32),
+  new MeshToonMaterial({
+    color: '#82DBC5',
+  }),
+)
+
```
andretchen0, line 65: In `nodeOps.ts`, I mentioned that cloning means that setup references are lost. Below is some demonstration code.

Both `torus` and `torusKnot` continue to exist after the `clone`, but they no longer point to the on-screen objects, so the on-screen objects don't update.

If the objects aren't `clone()`d in `nodeOps`, the on-screen shapes will rotate – but other errors pop up.

```
useRenderLoop().onLoop(() => {
  torus.rotateX(0.01)
  torusKnot.rotateY(0.01)
})
```

I don't know if there's a workaround here. Just pointing it out.

andretchen0 (changes requested):
I don't know if it can be worked around, but cloning the Three objects means that references in `setup` are lost. Details in the comments.

src/core/nodeOps.ts
alvarosabu: @andretchen0 Thats a good point thanks for raising it, the problem with the reference comes when we try to make the `:object` reactive to swap the instance, which completely changes the ref.

Not sure how to deal with that.

playground/src/pages/primitives.vue
alvarosabu: @andretchen0 This argument makes me think we should abandon the idea of trying to make `:object` prop reactive, swiping instances without losing the ref is getting too complicated.

Users could still use primitives with conditional rendering, I wanted `primitive` to work similar to `component` but honestly at this point I don't know how to achieve it.

andretchen0: @alvarosabu

I don't understand the internals of Tres well enough to have an opinion on the feasibility here. 

I'll start reading the source after I finish `<AnimatedSprite />` for Cientos.

>  I wanted primitive to work similar to component but honestly at this point I don't know how to achieve it.

I agree that this is a worthwhile goal. It'd be great if it "just works".

andretchen0: @alvarosabu

Sure thing. That could be fun!

andretchen0: Thanks! I'll go read the Discord.

src/core/nodeOps.ts
```diff
@@ -180,15 +182,41 @@ export const nodeOps: RendererOptions<TresObject, TresObject | null> = {
...
 
-    invalidateInstance(node as TresObject)
-    node.dispose?.()
   },
-  patchProp(node, prop, _prevValue, nextValue) {
+  patchProp(node, prop, prevValue, nextValue) {
     if (node) {
       let root = node
       let key = prop
+      if (key === 'object' && prevValue !== null) {
```
andretchen0, line 195: Is there ever another prop called `object` that will be passed here? Does this need to be more specific, e.g., check if the node is `primitive`?

alvarosabu: Good catch

alvarosabu: Done

src/core/nodeOps.ts
```diff
@@ -180,15 +182,41 @@ export const nodeOps: RendererOptions<TresObject, TresObject | null> = {
...
-    node.dispose?.()
   },
-  patchProp(node, prop, _prevValue, nextValue) {
+  patchProp(node, prop, prevValue, nextValue) {
     if (node) {
       let root = node
       let key = prop
+      if (node.__tres.primitive && key === 'object' && prevValue !== null) {
+        // If the prop 'object' is changed, we need to re-instance the object and swap the old one with the new one
+        const newInstance = nodeOps.createElement('primitive', undefined, undefined, {
```
andretchen0, line 197: It seems like this doesn't break references in `<script>` per se. The `:object`s still exist. But they're copied. That means something like the following (probably) wouldn't have the user's desired effect in the playground example:

```
useRenderLoop().onLoop(({elapsed}) => {
  torus.position.y = Math.sin(elapsed)
})
```

Whenever the primitive object becomes `torus`, `torus` is copied, along with its attributes. But the copy doesn't move like `torus` moves, because it's a different `Torus`.

andretchen0 (commented):
Again, I'm not sure if this can be worked around, but there's still a problem with references in `<script>`.

src/core/nodeOps.ts
alvarosabu: @andretchen0 we will need to explicitly mention on the docs that the user needs to animate the `ref` (primitiveRef) but not `the object` (torus) following the same DX as they were using a `<TorusMesh ref="torusRef" />` 

https://github.com/Tresjs/tres/assets/4699008/650d4633-c394-4a41-8c63-0b668cd878c5

There are several reasons why we need to `clone` the object:
- Since we soft-swipe the instances when `:object` changes (we maintain the reference but copy all the attributes) that would modify the original object from the setup leading to unwanted side effects.
It is not possible to pass a `ref` or `reactive` as @JaimeTorrealba mentioned in the thread. With clone, we can ensure that swiping between models is possible.

This is what happens without the `clone` 

https://github.com/Tresjs/tres/assets/4699008/b7ce86cc-44a1-4985-a91a-94b81df9c256

It's either has this limitation of having to use the ref coming from the custom renderer to animate or `:object` can't be reactive and users will need to use `v-if` and `v-else` to swipe models:

```
<primitive v-if="showA" :object="modelA" />
<primitive v-else :object="modelB" />
```

I balance through the first option.

andretchen0: @alvarosabu 

I think the PR as it currently stands is an improvement. I don't want to stand in the way of improvements.

andretchen0: This sort of error is an itch I can't stop scratching, so if you want to keep scratching ...

> Since we soft-swipe the instances when :object changes (we maintain the reference but copy all the attributes) that would modify the original object from the setup leading to unwanted side effects.

Can you elaborate on unwanted side effects?

Technical issues aside, there's a point that isn't clear for me conceptually. To illustrate:

```vue
<script>
...
const userMesh = new Mesh(...)
userMesh.position.x = 100

const userMesh2 = new Mesh(...)
userMesh2.position.x = 200
</script>

<template>
...
<primitive ref="primitiveRef" :object="userMesh" :position-x="300" />
...
</template>
```

* Should `userMesh`'s x be 300? (300 is the value in the `primitive`)
* And later, if the user does `primitiveRef.value.object = userMesh2`, does that set `userMesh2`'s x to 300?

For R3F, the answer to both is "yes", it seems. [Here's a StackBlitz.](https://stackblitz.com/edit/vitejs-vite-ggmpvs?file=src%2Fcomponents%2FPrimitiveTest.tsx) Notice that the rotation of the TorusRing increments with every frame, but resets whenever it's made the `primitive.object`. And the Torus would normally be placed at x = 2000, but its x is set to the primitive's value.

That seems like the right solution to me.

andretchen0: https://github.com/Tresjs/tres/blob/8408351e870a3e3cd530abd5b7ca58359c38f3e2/src/core/nodeOps.ts#L197

I understand that this will make a clone, which is required at the moment, but is there a reason why it's necessary to go through `nodeOps.createElement`? Is it updating something on the Vue side, maybe?

andretchen0 (approved):

src/core/nodeOps.ts
andretchen0: It seems like we're running into the "is-a" vs. "has-a" debate with Tres, though JS-style instead of classical OOP.

As it stands, the `node` in `nodeOps.ts` **is a** `TresObject`. Most of the time, it's a `THREE.Object3D` with some extra fields for Vue/Tres data. 

It's usually the case that `has-a` relationships have more flexibility. So, maybe we try that here with something like this shape for our node?

```
{
    tres: {
        // our Tres data
    },
    three: {
        // whatever THREE.Object3D, etc. is connected with this node – or none!
    }
}
```

Because `node` just has an optional reference to a `THREE` object, this has a nice knock-on effect in that we could have `<primitive />` as a placeholder without an `:object`.

I'll give it a shot and see how many errors that makes. ;)

PR #522, opened 2024-01-23
feat: 516 localstate for custom renderer node instances instead of userdata

### LINKED ISSUE #516: LocalState for custom renderer node instances instead of userData
### Description

As the author and maintainer of TresJS, I would like to simplify and clean the custom renderer `nodeOps` to make the code more readable and easier to maintain.

At, we are using overpopulation the property [userData](https://threejs.org/docs/#api/en/core/Object3D.userData) for the scene and objects on the scene graph as a workaround for accessing state or adding extra properties and methods that we need for TresJS nodeOps like this:

` scene.userData.tres__registerAtPointerEventHandler = registerObject`

`userData` is a property that the end-user can also use, we need to add the `.tres__` prefix to avoid naming collisions, adding an extra cognitive load to the code.

Code ends up being kind of difficult to read:

```ts
remove(node) {
    if (!node) return
    // remove is only called on the node being removed and not on child nodes.

    if (node.isObject3D) {
      const object3D = node as unknown as Object3D

      const disposeMaterialsAndGeometries = (object3D: Object3D) => {
        const tresObject3D = object3D as TresObject3D

        if (!object3D.userData.tres__materialViaProp) {
          tresObject3D.material?.dispose()
          tresObject3D.material = undefined
        }

        if (!object3D.userData.tres__geometryViaProp) {
          tresObject3D.geometry?.dispose()
          tresObject3D.geometry = undefined
        }
      }

      const deregisterAtPointerEventHandler = scene?.userData.tres__deregisterAtPointerEventHandler
      const deregisterBlockingObjectAtPointerEventHandler
        = scene?.userData.tres__deregisterBlockingObjectAtPointerEventHandler

      const deregisterAtPointerEventHandlerIfRequired = (object: TresObject) => {

        if (!deregisterBlockingObjectAtPointerEventHandler)
          throw 'could not find tres__deregisterBlockingObjectAtPointerEventHandler on scene\'s userData'

        scene?.userData.tres__deregisterBlockingObjectAtPointerEventHandler?.(object as Object3D)

        if (!deregisterAtPointerEventHandler)
          throw 'could not find tres__deregisterAtPointerEventHandler on scene\'s userData'

        if (
          object && supportedPointerEvents.some(eventName => object[eventName])
        )
          deregisterAtPointerEventHandler?.(object as Object3D)
      }

      const deregisterCameraIfRequired = (object: Object3D) => {
        const deregisterCamera = scene?.userData.tres__deregisterCamera

        if (!deregisterCamera)
          throw 'could not find tres__deregisterCamera on scene\'s userData'

        if ((object as Camera).isCamera)
          deregisterCamera?.(object as Camera)
      }

      node.removeFromParent?.()
      object3D.traverse((child: Object3D) => {
        disposeMaterialsAndGeometries(child)
        deregisterCameraIfRequired(child)
        deregisterAtPointerEventHandlerIfRequired?.(child as TresObject)
      })

      disposeMaterialsAndGeometries(object3D)
      deregisterCameraIfRequired(object3D)
      deregisterAtPointerEventHandlerIfRequired?.(object3D as TresObject)
    }

    node.dispose?.()
  },
 ``` 

### Suggested solution

Add a local state for each instance called `__tres` which contains:

```ts
export interface LocalState = {
  type: string
  // objects and parent are used when children are added with `attach` instead of being added to the Object3D scene graph
  objects: Instance[]
  parent: Instance | null
  primitive?: boolean
  eventCount: number
  handlers: Partial<EventHandlers>
  memoizedProps: { [key: string]: any }
}
```

```ts
export interface TresObject3D extends THREE.Object3D<THREE.Object3DEventMap> {
  geometry?: THREE.BufferGeometry & TresBaseObject
  material?: THREE.Material & TresBaseObject
  __tres: LocalState
}
```

We could include the context of TresContextProvider on the scene object to handle the register of cameras and event handlers or even inside each LocalState on a potential property `root`

```ts
export type LocalState = {
  type: string
  // objects and parent are used when children are added with `attach` instead of being added to the Object3D scene graph
  root: TresContext,
  objects: Instance[]
  parent: Instance | null
  primitive?: boolean
  eventCount: number
  handlers: Partial<EventHandlers>
  memoizedProps: { [key: string]: any }
}
```

### Alternative

_No response_

### Additional context

_No response_

### Validations

alvarosabu:
Closes #516

### DIFF
--- a/src/composables/useRaycaster/index.ts
+++ b/src/composables/useRaycaster/index.ts
@@ -20,10 +20,10 @@ interface PointerClickEventPayload {
 
 export const useRaycaster = (
   objects: Ref<THREE.Object3D[]>,
-  { renderer, camera, raycaster }: Pick<TresContext, 'renderer' | 'camera' | 'raycaster'>,
+  ctx: TresContext,
 ) => {
   // having a separate computed makes useElementBounding work
-  const canvas = computed(() => renderer.value.domElement as HTMLCanvasElement)
+  const canvas = computed(() => ctx.renderer.value.domElement as HTMLCanvasElement)
 
   const { x, y } = usePointer({ target: canvas })
 
@@ -39,11 +39,11 @@ export const useRaycaster = (
   }
 
   const getIntersectsByRelativePointerPosition = ({ x, y }: { x: number; y: number }) => {
-    if (!camera.value) return
+    if (!ctx.camera.value) return
 
-    raycaster.value.setFromCamera(new Vector2(x, y), camera.value)
+    ctx.raycaster.value.setFromCamera(new Vector2(x, y), ctx.camera.value)
 
-    return raycaster.value.intersectObjects(objects.value, false)
+    return ctx.raycaster.value.intersectObjects(objects.value, false)
   }
 
   const getIntersects = (event?: PointerEvent | MouseEvent) => {
--- a/src/core/nodeOps.ts
+++ b/src/core/nodeOps.ts
@@ -107,10 +107,7 @@ export const nodeOps: RendererOptions<TresObject, TresObject | null> = {
     if (child?.isObject3D) {
 
       if (child?.isCamera) {
-        if (!scene?.userData.tres__registerCamera)
-          throw 'could not find tres__registerCamera on scene\'s userData'
-
-        scene?.userData.tres__registerCamera?.(child as unknown as Camera)
+        child.__tres.root.registerCamera?.(child as unknown as Camera)
       }
 
       if (
@@ -179,7 +176,7 @@ export const nodeOps: RendererOptions<TresObject, TresObject | null> = {
       }
 
       const deregisterCameraIfRequired = (object: Object3D) => {
-        const deregisterCamera = scene?.userData.tres__deregisterCamera
+        const deregisterCamera = node.__tres.root.deregisterCamera
 
         if (!deregisterCamera)
           throw 'could not find tres__deregisterCamera on scene\'s userData'
--- a/src/core/nodeOps.ts
+++ b/src/core/nodeOps.ts
@@ -23,6 +23,17 @@ const supportedPointerEvents = [
   'onPointerLeave',
 ]
 
+export function invalidateInstance(instance: TresObject) {
+  const ctx = instance.userData.tres__root?.userData?.tres__context
+  
+  if (!ctx) return
+  
+  if (ctx.renderMode.value === 'on-demand' && ctx.internal.frames.value === 0) {
+    ctx.invalidate()
+  }
+
+}
+
 export const nodeOps: RendererOptions<TresObject, TresObject> = {
   createElement(tag, _isSVG, _anchor, props) {
     if (!props) props = {}
@@ -81,11 +92,17 @@ export const nodeOps: RendererOptions<TresObject, TresObject> = {
     return instance
   },
   insert(child, parent) {
-    if (parent && parent.isScene) scene = parent as unknown as TresScene
+    if (parent && parent.isScene) {
+      scene = parent as unknown as TresScene
+      if (child) {
+        child.userData.tres__root = scene
+      }
+    }
 
     const parentObject = parent || scene
 
     if (child?.isObject3D) {
+
       if (child?.isCamera) {
         if (!scene?.userData.tres__registerCamera)
           throw 'could not find tres__registerCamera on scene\'s userData'
@@ -180,6 +197,7 @@ export const nodeOps: RendererOptions<TresObject, TresObject> = {
       deregisterAtPointerEventHandlerIfRequired?.(object3D as TresObject)
     }
 
+    invalidateInstance(node as TresObject)
     node.dispose?.()
   },
   patchProp(node, prop, _prevValue, nextValue) {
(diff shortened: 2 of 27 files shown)


andretchen0 (commented):
@Tinoooo , can you handle this review?

I don't know enough about Tres internals to meaningfully participate here. (I'll start reading the code once Cientos' `<AnimatedSprite />` is finished.) 

In the abstract, I like the idea of moving away from `userData` so that users don't trip over us, so the changes seem worthwhile.

src/core/nodeOps.ts
```diff
@@ -74,49 +75,46 @@ export const nodeOps: RendererOptions<TresObject, TresObject> = {
       else if (instance.isBufferGeometry) instance.attach = 'geometry'
     }
 
+    instance.__tres = {
+      ...instance.__tres,
+      type: name,
+      memoizedProps: props,
+      eventCount: 0,
```
Tinoooo, line 82: **question:** What are `memoizedProps` and `eventCount` used for? I can't find them being used anywhere.

src/core/nodeOps.ts
```diff
@@ -74,49 +75,46 @@ export const nodeOps: RendererOptions<TresObject, TresObject> = {
...
-      if (props?.geometry?.isBufferGeometry) (instance as TresObject3D).userData.tres__geometryViaProp = true
-    }
-
-    // Since THREE instances properties are not consistent, (Orbit Controls doesn't have a `type` property) 
-    // we take the tag name and we save it on the userData for later use in the re-instancing process.
-    instance.userData = {
-      ...instance.userData,
-      tres__name: name,
+    if (instance.isObject3D && (props?.material || props?.geometry)) {
+      instance.__tres.disposable = false
```
Tinoooo, line 91: **question:** Does this work in cases when the material changes from being passed via prop to the default way? I'm not sure whether this is even supported though 😁.

src/core/nodeOps.ts
```diff
@@ -136,65 +134,52 @@ export const nodeOps: RendererOptions<TresObject, TresObject> = {
   },
   remove(node) {
     if (!node) return
+    const ctx = node.__tres
```
Tinoooo, line 137: **suggestion:** It would be nice if `instance.__tres` was typed.

Tinoooo:
> @Tinoooo , can you handle this review?
sure 😊

src/core/nodeOps.ts
alvarosabu: I added `memoizedProps` to be used on disposal and HMR cases where keeping the original `args` prop is necessary to re-construct the instance. Is not used here but its needed on #466 and maybe #514 

I added `eventCount` to be used for #529

src/composables/useRaycaster/index.ts
```diff
@@ -20,10 +20,10 @@ interface PointerClickEventPayload {
 
 export const useRaycaster = (
   objects: Ref<THREE.Object3D[]>,
-  { renderer, camera, raycaster }: Pick<TresContext, 'renderer' | 'camera' | 'raycaster'>,
+  ctx: TresContext,
```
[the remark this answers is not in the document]
alvarosabu: Hey Tino, thanks for the feedback. While I understand the reasoning behind being verbose about the deps, don't you consider it overcomplicated when you read it through? 

> because some parts of the context might not exist yet.

deconstructing needs the same check as `ctx.camera === undefined` in case it doesn't exist yet.

In this specific instance, I opted for passing the entire TresContext (ctx) for a few reasons:

1. **Simplicity**: The function in question utilizes multiple properties of `TresContext`. Passing the entire context rather than destructuring simplifies the function signature.

2. **Future-proofing**: If we need to use more properties of `TresContext` in the future, we can avoid modifying the function signature each time, reducing the potential for errors and merge conflicts.

3. **Performance Consideration**: Yep, it's not a significant impact in most cases, avoiding destructuring means we don't create additional variables unnecessarily, which is a marginal optimization.

We need also to consider what would be easier to maintain, especially if we want to delegate some level of maintenance to the community.

src/core/nodeOps.ts
alvarosabu: Good question, need to try it out, what would you suggest, to add an else and set disposable to true again?

src/core/nodeOps.ts
alvarosabu: Check https://github.com/Tresjs/tres/pull/522/files/aaa4d322d3e049c1ebd8b1f0362bca464471067e#r1472657228

src/types/index.ts
```diff
@@ -37,29 +37,30 @@ interface TresBaseObject {
...
 // Custom type for geometry and material properties in Object3D
 export interface TresObject3D extends THREE.Object3D<THREE.Object3DEventMap> {
   geometry?: THREE.BufferGeometry & TresBaseObject
   material?: THREE.Material & TresBaseObject
-  userData: {
-    tres__materialViaProp: boolean
-    tres__geometryViaProp: boolean
-    [key: string]: any
-  }
+  __tres: LocalState
```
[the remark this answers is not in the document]
alvarosabu: Ohhh, I manage to solve it, forget about it:

```ts
export type TresObject = 
  TresBaseObject & (TresObject3D | THREE.BufferGeometry | THREE.Material | THREE.Fog) & { __tres: LocalState }

```

PR #529, opened 2024-01-31
feat(events)!: pointerevents manager and state

garrlker:
### Events
- [x] onClick
- [x] onContextMenu (rightClick)
- [x] onDoubleClick
- [x] onWheel
- [x] onPointerDown
- [x] onPointerUp
- [x] onPointerLeave
- [x] onPointerMove
- [x] onPointerMissed

### Event Details
- [x] ...DomEvent
- [x] ...Intersection
- [x] intersections: Intersection[]
- [x] object: Object3D
- [x] eventObject: Object3D
- [ ] unprojectedPoint: Vector3
  - halfway done, need to clean up some math
- [x] ray: Ray
- [x] camera: Camera
- [x] sourceEvent: DomEvent
- [x] delta: number

### Features

- [x] Event support for Primitives
- [x] Event Bubbling and Propagation
	- [x] Raycast Propogation
	- [x] Hierarchy Propogation
- [x] Events are emitted off of `TresCanvas`
- [x] Forced Raycasts
- [x] Duplicate Checks (no event handler called twice)

### Chores
- [x] Getting up to date with the v4 branch
- [x] Documentation
- [ ] Unit tests


playground/src/pages/raycaster/Propogation.vue
alvarosabu, line 1: typo `Propagation.vue`

src/composables/useRaycaster/index.ts
```diff
@@ -43,10 +48,10 @@ export const useRaycaster = (
 
     raycaster.value.setFromCamera(new Vector2(x, y), camera.value)
 
-    return raycaster.value.intersectObjects(objects.value, false)
+    return raycaster.value.intersectObjects(objects.value, true)
```
andretchen0, line 51: If I understand correctly, this will test against all children of the scene and recurse down through all children.

If that's right, is that worth optimizing in this PR? Or ever?

----

For context, it looks like R3F does [some optimization here to keep some eventless objects](https://github.com/pmndrs/react-three-fiber/blob/bc0d4023a256d2cb0e04450d0803ceb3aa3928f4/packages/fiber/tests/core/renderer.test.tsx#L392) from being tested for intersections.

andretchen0 (changes requested):
@garrlker 

Looks like events are coming along! Great!

I'm not overly familiar with Tres internals, so this review is taking me a while. I'll have another look tomorrow, but I'll go ahead and submit what I've got so you can take a look when it's convenient.

playground/src/pages/raycaster/Propogation.vue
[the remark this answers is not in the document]
garrlker: I didn't even consider this use case, and you are 100% right! Great catch!


Removing the `createGlobalState`  let's it work for both canvas' on my system. Thanks for the reproduction!

src/composables/useRaycaster/index.ts
andretchen0: Sounds good! 

It sounds like there's plenty to think about and discuss with the others. I don't want to keep these events out of Tres. So maybe we open another PR when this PR is merged.

> Only include node's with events and their direct descendants in the hit test. Being a custom renderer gives us this flexibility to do this on scene insertion.

Yeah, something like that is what I had in mind as well. We'd have to watch event updates as well.

src/composables/useEventStore/index.ts
```diff
@@ -0,0 +1,101 @@
...
+          intersection.stopPropagation = () => stopPropagating = true;
+
+          const { object } = intersection
+          executeEventListeners(object[eventName], intersection, event)
+
+          // Propogate the event up the parent chain before moving on to the next intersected object
+          let parent = object.parent
+          while(parent !== null && !stopPropagating) {
+            executeEventListeners(parent[eventName], intersection, event)
+            parent = parent.parent
```
andretchen0, line 58: ## One click should probably be no more than one call to a click handler.

For a given click, etc., a particular click handler shouldn't be called more than once, it seems to me.

## Currently

This shows the result of a single click with the current bubbling setup:

Screenshot 2024-02-02 at 02 42 39

The console shows that the box at the top of the pyramid gets its click handler called 4 times.

## "Realer" world

The above doesn't seem wrong to me, but I was trying to think about it as a Tres user and I think it breaks down in "realer" world cases.

Here's a simple example: buy a bicycle.

```vue
<Bicycle @click="buy">
  <Frame>
    <Wheel />
    <Wheel />
  </Frame>
</Bicycle>
```

Positioning the canvas at a certain angle, you could click both wheels and the frame, and you'll buy 3 bikes instead of 1.

## R3F removes duplicates

I checked out R3F. [They hey remove duplicates.](https://github.com/pmndrs/react-three-fiber/blob/0ef66a1d23bf16ecd457dde92b0517ceec9861c5/packages/fiber/src/core/events.ts#L191C5-L191C41) 

Here's a [StackBlitz with a simple setup like your playground](https://stackblitz.com/edit/vitejs-vite-ggmpvs?file=src%2FApp.tsx,src%2Fcomponents%2FBox.tsx).

playground/src/pages/raycaster/Propogation.vue
garrlker: Ty, fixed

src/composables/useRaycaster/index.ts
garrlker: After talking with @alvarosabu and @Tinoooo it looks like we have enough time before the V4 release for me to take a stab at optimizing this some more

I was able to confirm `intersectObjects` is being called multiple times per frame (and even per event).
I've simplified it's usage. I'm seeing about half the cpu % in my test scenes.

alvarosabu: I think we could add a flag to the localstate `__tres` of the instances marking if we expect it to have events to optimize the traversing of the scene. See recently merged #522 

Meshes and groups:

```
instance.__tres.events = true
```

Dunno, materials and geometries:

```
instance.__tres.events = false
```

alvarosabu:
@garrlker Regarding the store, I would suggest adding it to the `TresContext` directly instead of creating a separate store, that way is available on all instances via localState

src/composables/useRaycaster/index.ts
garrlker: Ahh good idea. The localstate `__tres` will be very helpful

Thank you for calling it out!

src/composables/useEventStore/index.ts
garrlker: Great catch!

I will be sure to add duplicate checks somehow

andretchen0:
@alvarosabu 

> Regarding the store

As an aside, is there any interest in using something like Pinia for the store?

playground/src/pages/raycaster/Propagation.vue
andretchen0, line 1: Looks like a single click can still trigger a handler multiple times:

Screenshot 2024-03-09 at 15 18 39 copy

## Context

This [earlier comment](https://github.com/Tresjs/tres/pull/529#discussion_r1475415652) has details.

andretchen0:
> @garrlker Regarding the store, I would suggest adding it to the `TresContext` directly instead of creating a separate store, that way is available on all instances via localState

@alvarosabu , that answers this question from the @garrlker 's source, right?

>  // QUESTION: Maybe I should move the useEventStore logic into usePointerEventHandler? I seem to have created a similar pattern
>  // usePointerEventHandler({ scene: scene.value, contextParts: context.value })

src/composables/useEventStore/index.ts
```diff
@@ -0,0 +1,104 @@
...
+     * Propogates an event to all intersected objects and their parents
+     * @param eventName - The name of the event to propogate
+     * @param event - The event object
+     * @param intersects - An array of intersections
+     */
+  function propogateEvent(eventName: string, event, intersects) {
+    let stopPropagating = false
+    const stopPropagatingFn = () => stopPropagating = true
+    // Loop through all intersected objects and call their event handler
+    if (intersects.length) {
```
andretchen0, line 42: (Non-blocking)

If you know `intersects` is an array, this conditional can be removed. `for (const v of vs)` won't loop if `vs.length === 0`. So this conditional is already baked into the next line.

andretchen0 (changes requested):

Tinoooo:
> @alvarosabu 
> 
> 
> 
> > Regarding the store
> 
> 
> 
> As an aside, is there any interest in using something like Pinia for the store/`context`? Probably after v4?
> 
> 
> 
> ### Context
> 
> 
> 
> [R3F uses their own reactive Pinia-like store (zustand) for similar purposes.](https://github.com/pmndrs/react-three-fiber/blob/e1acae8e1aea6ac8b255d42ffd19d0bfdd6cd529/packages/fiber/src/web/events.ts#L1)

The context provider pattern (inspired by @maoberlehner s [blog article](https://markus.oberlehner.net/blog/context-and-provider-pattern-with-the-vue-3-composition-api/) btw.) we currently have allows us to have a certain context for each TresCanvas in an app. I have the feeling, that a pinia store is directed to be used as a global store (meaning it scopes throughout the whole vue app). I am not 100% sure about this though. There is probably also a way to have it scoped. What advantages do you see if we were using pinia @andretchen0 ?

alvarosabu:
Hey @andretchen0 @Tinoooo love the discussion. In the beginning, we evaluated Pinia, but we decided to drop it because we felt it was overengineered for what we needed and it adds an extra dependecy to the project (I personally prefer to keep external deps as few as possible). 

For our use case, a small reactive state was more than enough, but I'm open to knowing what benefit would we get from Pinia.

andretchen0:
@Tinoooo @alvarosabu 

I don't have much experience with Pinia, so thanks for the input. I'll drop it as "unsuited for our needs". 

You're right that it appears that Pinia doesn't offer out-of-the-box per-instance stores – a bit of a bummer. Though maybe I haven't found the correct spot in the documentation.

> What advantages do you see if we were using Pinia @andretchen0 ?

One positive aspect of some opinionated stores is that they *can* help enforce a style where store users only interact with functions: getters and actions. Users can read all exposed data (getters), but only modify it in ways the store owner provides (actions).

I find that to be a helpful paradigm.

Tinoooo:
@andretchen0 I agree with you. Some parts (cameras, and the computeds) should not be writable.  We currently use [readonly](https://vuejs.org/api/reactivity-core.html#readonly) to achieve that.

alvarosabu:
> Stores can help enforce a style where store users only interact with functions: getters and actions. 

I also agree, we currently design it to have readonly and computed properties, normally Pinia offers also things like type safety, devtools and such but that's something we are providing from our side.

src/composables/useEventStore/index.ts
garrlker: good catch! Ty, it is removed

src/composables/useRaycaster/index.ts
garrlker: I've added the EventManager onto Tres's Context

Let me know if that was what you had in mind

src/composables/useEventStore/index.ts
garrlker: The latest changes include duplicate checks and it's working pretty well

Thanks for catching that!

playground/src/pages/raycaster/Propagation.vue
garrlker: Fixed in latest, but please test to ensure it meets your expectations

garrlker:
Hey everyone! I've pushed up some new changes that should hopefully resolve the current comments

They include event duplicate checks, onPointerMissed support, and forced event updating (For cases when the mouse hasn't moved)

I will update this list later but as of right now this PR likes

- [ ] Getting up to date with the v4 branch
- [ ] Documentation
- [ ] Unit tests
- [ ] More testing(from you guys)
- [ ] Are we missing any features we wanting to implement?

I'm going to push the perf optimization work into a new issue for later so it won't hold this up. The perf is still better than before so I don't think it's high priority.

src/composables/useTresContextProvider/index.ts
```diff
@@ -45,6 +45,20 @@ export interface PerformanceState {
   }
 }
 
+interface TresEventManager {
```
andretchen0, line 48: (Non-blocking)

At first glance, it seems like this should be in `src/componsables/useTresEventManager/index.ts`.

andretchen0:
@garrlker 

Looks good so far!

Maybe out of scope for this PR, but how easy is it for a user to replace this event system with another? Is there a way for a user to do that without modifying the core?

playground/src/pages/raycaster/Propagation.vue
andretchen0: Gave it an initial test and it seems like the duplicate calls are gone. Great! 👍

garrlker:
As of right now there's no way to replace it without modifying the core but I think we could support that long term and not need too many changes

alvarosabu:
Ohhh @garrlker this already looks great. I just tested the demos and the different options, propagation, and especially` ev.stopPropagation()` are working as I would expect them. Great job!

I think we can improve the Event object that we are returning. Comparing quickly the event returned by clicking (left R3F vs right Vue):

Screenshot 2024-03-19 at 16 54 17

For example, returning and adding type of Event would be important:

```ts
{
  type: "click",
  distance: ...
}
```
Screenshot 2024-03-19 at 16 59 21

For example here, Im truing to stop the propagation of the `Boxes` on the playground, since the color was pass as parameter now I would need to decide the color programmatically inside of the click, so having info about the type of pointer could be good

andretchen0:
> wdyt @garrlker , @andretchen0 do you see any other property that could be important for the end user?

Here's the R3F approach in a nutshell:

> Events contain the browser event as well as the three.js event data (object, point, distance, etc).

https://docs.pmnd.rs/react-three-fiber/api/events

That seems like a good approach to me.

garrlker:
Great catch!

ngl I've never liked how verbose R3F events can be, mostly because I like to log the whole events in my console when debugging. But that's more personal preference 😅

I've started working on adding more properties to the event details object.

Also, I've added my personal todo list to the PR so it's easier to see what's been implemented. If you think anythings missing feel free to add more.

alvarosabu:
Hey @garrlker I just updated the branch to latest v4, everything smooth,I'm also gonna take over docs

- [x]  Getting up to date with the v4 branch
- [ ]  Documentation

Regarding:

> I'm going to push the perf optimization work into a new issue for later so it won't hold this up. The perf is still better than before so I don't think it's high priority.

We can iterate in a follow up task after `v4`

alvarosabu:
@garrlker Doing testing of each event I realise that the `pointer-missed` returns an empty array 

Screenshot 2024-03-22 at 10 33 48

I checked it on R3F and you get an event object similar to the other events with the type of event that you missed. Could we have a similar behavior?

andretchen0:
Tested with primitives. Works! Great! 👍

alvarosabu:
@andretchen0 that was my next test, thanks for taking care of it 💚

andretchen0:
> @andretchen0 that was my next test, thanks for taking care of it 💚

No problem. 😉

<details>
<summary>
Fwiw, here's the Vue file I set up if that's handy for anyone for double-checking primitives.

</summary>

```vue
<script setup lang="ts">
import type { ThreeEvent } from '@tresjs/core'
import { BasicShadowMap, SRGBColorSpace, NoToneMapping, BoxGeometry, MeshToonMaterial, Mesh, MathUtils } from 'three'
import { TresLeches, useControls } from '@tresjs/leches'
import { OrbitControls } from '@tresjs/cientos'
import '@tresjs/leches/styles'

const gl = {
  clearColor: '#202020',
  shadows: true,
  alpha: false,
  shadowMapType: BasicShadowMap,
  outputColorSpace: SRGBColorSpace,
  toneMapping: NoToneMapping,
}

const { stopPropagation } = useControls({
  stopPropagation: false,
})

function onClick(ev: ThreeEvent<MouseEvent>) {
  console.log('click', ev)
  if (stopPropagation.value) ev.stopPropagation()
  ev.object.material.color.set('#008080')
}

function onDoubleClick(ev: ThreeEvent<MouseEvent>) {
  console.log('double-click', ev)
  if (stopPropagation.value) ev.stopPropagation()
  ev.object.material.color.set('#FFD700')
}

function onPointerEnter(ev: ThreeEvent<MouseEvent>) {
  if (stopPropagation.value) ev.stopPropagation()
  ev.object.material.color.set('#CCFF03')
}

function onPointerLeave(ev: ThreeEvent<MouseEvent>) {
  if (stopPropagation.value) ev.stopPropagation()
  /*  ev.object.material.color.set('#efefef') */
}

function onPointerMove(ev: ThreeEvent<MouseEvent>) {
  if (stopPropagation.value) ev.stopPropagation()
}

function onContextMenu(ev: ThreeEvent<MouseEvent>) {
  console.log('context-menu', ev)
  if (stopPropagation.value) ev.stopPropagation()
  ev.object.material.color.set('#FF4500')
}

function onPointerMissed(ev: ThreeEvent<MouseEvent>) {
  console.log('pointer-missed', ev)
  if (stopPropagation.value) ev.stopPropagation()
}

const SIDE = 3
const COUNT = SIDE * SIDE * SIDE
const SPREAD = 6
const [LO, HI] = [-SPREAD * 0.5, SPREAD * 0.5]
const lerp = MathUtils.lerp

const geo = new BoxGeometry()
const objs = new Array(COUNT).fill(0).map((_, i) => {
  const mesh = new Mesh(geo, new MeshToonMaterial({ color: '#efefef' }))
  mesh.position.set(
    lerp(LO, HI, (i % SIDE) / (SIDE - 1)), 
    lerp(LO, HI, (Math.floor(i / SIDE) % SIDE) / (SIDE - 1)),
    lerp(LO, HI, Math.floor(i / (SIDE * SIDE)) / (SIDE - 1)),
  )
  return mesh
})
</script>

<template>
  <TresLeches />
  <TresCanvas
    window-size
    v-bind="gl"
  >
    <TresPerspectiveCamera :position="[10, 10, 10]" />
    <OrbitControls />
    <primitive
      v-for="obj, i in objs"
      :key="i"
      :object="obj"
      @click="onClick"
      @double-click="onDoubleClick"
      @pointer-enter="onPointerEnter"
      @pointer-leave="onPointerLeave"
      @pointer-move="onPointerMove"
      @context-menu="onContextMenu"
      @pointer-missed="onPointerMissed"
    />
    <TresDirectionalLight :intensity="1" />
    <TresAmbientLight :intensity="1" />
  </TresCanvas>
</template>

```
</details>

garrlker:
Hey everyone, I've pushed up my latest. This includes a few fixes from the previous comments and some other features

Most notably it includes a much more detailed events object for all events, including the `pointer-missed` event.

Besides that it also includes the ability to force a raycast(events fire even when mouse isn't moving) and I fixed a major oversight I came across in the pointer-enter/leave events

This should be good for more testing

alvarosabu (approved):
Amazing work @garrlker 👏🏻 we can finally merge it.

PR #530, opened 2024-01-31
feat(app): Add a new directive, v-rotate

JaimeTorrealba:
A simple new idea for a directive. Let me know what you think

Also fix some typos in the docs of the other directives


src/directives/vRotate.ts
```diff
@@ -0,0 +1,61 @@
...
+  mounted: (
+    el: TresObject,
+    binding: {
+      arg: 'x' | 'y' | 'z'
+      value: number
+      modifiers: Partial<{ x: boolean; y: boolean; z: boolean }>
+    },
+  ) => {
+    if (el.isCamera) {
+      logWarning(`Rotate the ${el.type} is not a good idea`)
```
andretchen0, line 18: That's really thoughtful of you! 🤣

andretchen0:
```ts
if (el.isCamera) {
      logWarning(`Rotate the ${el.type} is not a good idea`)
```

That's really thoughtful of you! 🤣

src/directives/vRotate.ts
```diff
@@ -0,0 +1,61 @@
...
+      return
+    }
+    const speed = binding.value ?? 0.01
```
andretchen0, line 21: (Non-blocking)

I personally find it really helpful if this sort of variable is named using units. Here, that would be `radiansPerFrame` (on a single axis), if I'm understanding the code correctly.

Something like `speed` is fine if I just want to tweak a value until it looks right. But particularly for cases where I want to sync the movement of two different components, or where I might want to include an equation from a physics textbook, it's *really* handy to have the units in the variable name.

src/directives/vRotate.ts
```diff
@@ -0,0 +1,61 @@
...
+    const quaternionY = shallowRef(
+      binding.modifiers.y || binding.arg === 'y'
+        ? new Quaternion().setFromAxisAngle(new Vector3(0, 1, 0), speed)
+        : defaultQuaternion,
+    )
+    const quaternionZ = shallowRef(
+      binding.modifiers.z || binding.arg === 'z'
+        ? new Quaternion().setFromAxisAngle(new Vector3(0, 0, 1), speed)
+        : defaultQuaternion,
+    )
```
andretchen0, line 37: (Non-blocking)

You could get by with a single quaternion here, rather than using a quaternion per axis. Something like this pseudocode:

```
new Quaternion().setFromAxisAngle(new Vector3(x, y, z).normalize(), speed)
```

That would have the positive (imho) side effect of maintaining a consistent rotational speed, regardless of arguments, for a given frame rate. 

E.g., as it stands, if the user passes `x` and `y`, the object rotates around the x-axis at `speedPerFrame` and around the y-axis at `speedPerFrame`. With normalization, the sum of all rotations will be a consistent `speedPerFrame`.

src/directives/vRotate.ts
```diff
@@ -0,0 +1,61 @@
...
+      )
+    }
+
+    const { onLoop } = useRenderLoop()
+
+    onLoop(() => {
+      el.applyQuaternion(quaternionX.value)
+      el.applyQuaternion(quaternionY.value)
+      el.applyQuaternion(quaternionZ.value)
+    })
```
andretchen0, line 59: (Non-blocking)

This code ...

```
    onLoop(() => {
      el.applyQuaternion(quaternionX.value)
```

... ties the rotational speed to the frame rate. E.g., if the frame rate drops, the object will rotate more slowly.

To maintain consistency across frame rates, it's probably best to [make movement/rotation a function of `delta` time](https://dev.to/dsaghliani/understanding-delta-time-in-games-3olf).

andretchen0 (commented):
Hey @JaimeTorrealba !

This could be handy! 👍

I made some non-blocking comments.

src/directives/vRotate.ts
JaimeTorrealba: Hi I tried this, but I get weird behaviors, the cube rotate, but started to grow in every rotation...
So I decided to go with the other approach

src/directives/vRotate.ts
JaimeTorrealba: I like this one. Thanks :)

src/directives/vRotate.ts
JaimeTorrealba: I know about this, but I didn't want to create the Quaternionin every frame (as a const) or create a function.

This is probably directive, that you'll use for debug purpose, scenes (that's my original thought maybe I'm wrong)
I find myself rotating meshes in my scenes

src/directives/vRotate.ts
andretchen0: Like this?

https://github.com/Tresjs/tres/assets/20469369/3f6caa15-343c-430c-a758-3baf91ceec95

If so, it might be due to a non-normalized `Quaternion` or non-normalized axis `Vector3`. Here's the code that produced clip, for example:

```
useRenderLoop().onLoop(({elapsed}) => {
    if (boxRef.value) {
      boxRef.value.value.applyQuaternion(new Quaternion().setFromAxisAngle(new Vector3(1,1,1), 0.1))
    }
})
```

And here's the normalized version:

https://github.com/Tresjs/tres/assets/20469369/921a1163-2592-4320-8071-cf562dcec101


```
useRenderLoop().onLoop(({elapsed}) => {
    if (boxRef.value) {
      boxRef.value.value.applyQuaternion(new Quaternion().setFromAxisAngle(new Vector3(1,1,1).normalize(), 0.1))
    }
})
```

src/directives/vRotate.ts
andretchen0: No problem. Take whatever's useful and discard the rest. 🙂

src/directives/vRotate.ts
JaimeTorrealba: Ooohhh yess exactly like that, Thanks.

It was because is not normalize, thanks my friend ❤️

alvarosabu (changes requested):
Looks good @JaimeTorrealba please coordinate with the different owners of translations to translate the new addition to the docs. You can do that on the #contributions channel on discord

alvarosabu:
@JaimeTorrealba please remember using kebab-case on the branch names.

JaimeTorrealba:
Close for #651

PR #531, opened 2024-02-01
fix: camera aspect

andretchen0:
> [!WARNING]
> For current users of `<TresOrthographicCamera />` this is a breaking change. A solution is provided but requires intervention. See "Problem 2".

## Problem 1

https://github.com/Tresjs/tres/assets/20469369/9b6d8bbb-3508-43b6-9316-3f68ed5edbaf

Unlike `<TresPerspectiveCamera />`, `<TresOrthographicCamera />` does not update automatically after screen ratio changes. 

[StackBlitz](https://stackblitz.com/edit/tresjs-basic-wq1b5y?file=src%2Fcomponents%2FTheExperience.vue)

## Solution

R3F uses the `Three.OrthographicCamera`'s default `left, right, top, bottom` to maintain the frustum cuboid. This PR follows that example.

----

## Problem 2

Tres currently overwrites `aspect` on `<TresPerspectiveCamera />`. In rare cases, the user may wish to control `aspect`. 

Similarly, with this PR, `left, right, top, bottom` on `<TresOrthographicCamera />` are controlled by Tres. Users who set those attributes manually need a way to keep Tres from overwriting them.

## Solution

R3F lets users set a `manual` flag on the cameras to opt out of automatic aspect changes. This PR follows that example.

To set a camera as "manual" in TypeScript:

```ts
const myCamera = new PerspectiveCamera(); // or OrthographicCamera
(myCamera as any).manual = true
```

----

## Playground

Playground updated at /cameras/camera . `manual` cameras have been added.


alvarosabu:
Hey @andretchen0 great work here, thanks for the fix, if it's a breaking change, should we include it better on the v4 then?

andretchen0:
> Hey @andretchen0 great work here, thanks for the fix, if it's a breaking change, should we include it better on the v4 then?

Thanks! v4 is fine with me.

alvarosabu:
@andretchen0  Ok awesome, just change the base of the PR to the branch `v4`

alvarosabu:
@andretchen0 something is off, there are too many changes that weren't related to your initial PR into v4

andretchen0:
@alvarosabu 

Thanks for flagging that.

Merging in v4 seems to have added a lot of files to the PR. I'll close open a new PR that's based on v4 from the start.

PR #538, opened 2024-02-06
fix: camera aspect

andretchen0:
## Problem

Orthographic camera is not getting updates when screen size changes. So the aspect ratio falls out of sync with window size, resulting in squished/stretched images.

## Solution

* Update orthographic camera aspect when screen size updates
* Give user a "manual" flag to keep Tres from updating camera

## Context

[Manual flag in R3F.](https://github.com/pmndrs/react-three-fiber/blob/0ef66a1d23bf16ecd457dde92b0517ceec9861c5/packages/fiber/src/core/utils.ts#L457)

### DIFF
--- a/playground/src/pages/cameras/Cameras.vue
+++ b/playground/src/pages/cameras/Cameras.vue
@@ -13,16 +13,17 @@ const gl = {
   outputColorSpace: SRGBColorSpace,
   toneMapping: NoToneMapping,
 }
+type Cam = (PerspectiveCamera | OrthographicCamera) & { manual?: boolean }
 
 const state = reactive({
   cameraType: 'perspective',
-  camera: new PerspectiveCamera(75, 1, 0.1, 1000),
+  camera: new PerspectiveCamera(75, 1, 0.1, 1000) as Cam,
 })
 
 state.camera.position.set(5, 5, 5)
 state.camera.lookAt(0, 0, 0)
 
-const { value: cameraType } = useControls({
+const { cameraType, manual } = useControls({
   cameraType: {
     label: 'CameraType',
     options: [{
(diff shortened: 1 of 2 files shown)


alvarosabu (approved):
Thanks for this @andretchen0. I like the `manual` prop, I'm thinking how we could add it to the docs so the users are aware of it.

andretchen0:
> Thanks for this @andretchen0. I like the `manual` prop, I'm thinking how we could add it to the docs so the users are aware of it.

Maybe we add an entry in https://docs.tresjs.org/advanced/ ? 

I'm not sure many users will want it. It's pretty niche. Afaik, no one had complained that Tres was controlling the `<TresPerspectiveCamera />` previously. 

I just don't want to force people to drop Tres because their use case requires controlling the camera.

alvarosabu:
> > Thanks for this @andretchen0. I like the `manual` prop, I'm thinking how we could add it to the docs so the users are aware of it.
> 
> Maybe we add an entry in https://docs.tresjs.org/advanced/ ?
> 
> I'm not sure many users will want it. It's pretty niche. Afaik, no one had complained that Tres was controlling the `<TresPerspectiveCamera />` previously.
> 
> I just don't want to force people to drop Tres because their use case requires controlling the camera.

I would say you already found it valuable so let's not wait for the community to request it or complain about it, if someone asks, we can direct the user to that documentation.

Outside of the scope of this PR, @andretchen0 I'm strongly thinking about adding a cookbook to the docs similar to this https://docs.astro.build/en/recipes/., maybe is a good candidate?

alvarosabu:
@andretchen0 hey buddy, can you confirm me this would not need to update any docs? If not then we can merge it already

andretchen0:
It would potentially break an existing orthographic camera that the user was updating themselves — which was really the only choice, previously.

Maybe that needs documentation?

andretchen0:
> I hardly think anyone was doing that until now 😂, it's ok then

I think you're right. Lol

PR #549, opened 2024-02-09
chore: add lint:fix script

Dekier:


andretchen0:
## Strong preferences

* If we change this in `Tres`, we should do it all the subprojects, e.g., Cientos. (I bounce around quite a bit between them. 🤪)
* Whatever the `fix` incantation becomes, it should cover the same files as the regular lint.

## Weak preference

`pnpm run lint --fix` works fine for me across Tres and the subprojects. It also runs on the same files as `pnpm run lint`.

andretchen0:
Closing due to inactivity. Feel free to reopen.

PR #579, opened 2024-03-07
fix: `nodeOps` is now a function

### LINKED ISSUE #560: NodeOps shares `scene` across multiple`TresCanvas`s
## Bug

NodeOps accidentally shares `scene` across multiple `TresCanvas`s.

See the StackBlitz reproduction for details.

## Problem

NodeOps's [`let scene`](https://github.com/Tresjs/tres/blob/f6097a3d6a3e34fddb689266b062869fe5a9bf1e/src/core/nodeOps.ts#L15) is [set whenever a new `Scene` is added to any `TresCanvas`](https://github.com/Tresjs/tres/blob/f6097a3d6a3e34fddb689266b062869fe5a9bf1e/src/core/nodeOps.ts#L84) – i.e., the variable points to the last `Scene` added to a `TresCanvas`. If elements from another `TresCanvas` read `scene` in nodeOps, the code expects that they will find *their* `Scene`, but they do not.

### Solution A

We could refactor nodeOps as `(context: TresContext) => NodeOps`. 

Every new instance of `TresCanvas` could create a new nodeOps instance which contains a unique `TresContext` (i.e., with the appropriate `Scene`) within its scope.

### Solution B

Otherwise, we could simply recurse up the parent chain until we reach a `Scene` or run out of parents.

## Reproduction

https://stackblitz.com/edit/tresjs-basic-e6uzmm?file=src%2FApp.vue

## System Info

```shell
System:
    OS: Linux 5.0 undefined
    CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 18.18.0 - /usr/local/bin/node
    pnpm: 8.14.0 - /usr/local/bin/pnpm
  npmPackages:
    @tresjs/cientos: ^3.5.1 => 3.5.1 
    @tresjs/core: ^3.4.1 => 3.4.1 
    @tresjs/eslint-config-vue: ^0.2.1 => 0.2.1 
    vite: ^4.5.0 => 4.5.0
```

### Used Package Manager

npm

### Code of Conduct

alvarosabu:
Hopefully closes #560


src/core/nodeOps.ts
```diff
@@ -33,8 +32,9 @@ export function invalidateInstance(instance: TresObject) {
 
 }
 
-export const nodeOps: RendererOptions<TresObject, TresObject | null> = {
-  createElement(tag, _isSVG, _anchor, props): TresObject | null {
+export const nodeOps: () => RendererOptions<TresObject, TresObject | null> = () => {
```
andretchen0, line 35: (non-blocking)

Maybe use a verb now, since this is a function? Rename to `getNodeOps`?

src/components/TresCanvas.vue
```diff
@@ -87,6 +88,7 @@ const slots = defineSlots<{
 }>()
 
 const instance = getCurrentInstance()?.appContext.app
+extend(THREE)
```
andretchen0, line 91: (Non-blocking)

I think this is coupled with `nodeOps`, right? (Like, if we created a totally different Vue renderer, we wouldn't need this at all.)

So, can it be close to line 111 or else inside of `nodeOps`?

andretchen0 (approved):
Nice fix! ;)

Just a few comments. All non-blocking.

src/components/TresCanvas.vue
alvarosabu: > If we created a totally different Vue renderer for something besides THREE, we wouldn't need this at all.

Do we? If we would create a custom renderer for, dunno, Babylon, it should exist in a completely different package.

PR #583, opened 2024-03-09
feat(core): allow using window as pointer events source

### LINKED ISSUE #581: Use window for pointerevents
### Description

I have a TresCanvas with a lower z-index that is under the main content on my site. I got around this in native Three by using events based on window pointer events. I don't see a way to do this in Tres. 

### Suggested solution

Add a prop to TresCanvas like window-size (window-pointer?) to use window pointer events instead of attaching events to the canvas itself

### Alternative

_No response_

### Additional context

_No response_

### Validations

acvigue:
Added a `useWindowPointer` prop to TresCanvas that states whether or not to use the window as the source of pointer events.

Closes #581 

Added the prop to docs page, let me know if I'm missing anything 😄


andretchen0:
@alvarosabu @Tinoooo 

What do you think about expanding the `TresCanvas` props here?

https://github.com/Tresjs/tres/blob/43cd1a4a63fc58417ce90c472288883673315a5d/src/components/TresCanvas.vue#L65

My take: 

I'm completely on board with allowing additional ways for the system to get events. 

But, rather than passing a boolean flag that has to be implemented by the system for this particular case, I'd prefer to invert the dependencies. E.g., define a default `PointerEventSource` and allow the users to optionally provide their own.

Maybe that's something that can wait though.

alvarosabu:
Hi @acvigue the pR probably needs to be updated with the v4 pointer event changes, could you do it so we evaluate itadding it in the next iterations? cc @andretchen0

andretchen0:
Closing here.

For #789 , we'll use dependency inversion to allow users to supply their own pointer.

PR #599, opened 2024-03-25
fix(core): add new launch.json and fixed breakpoints not working

ThimoDEV:
This PR adds a new launch.json so a maintainer of TresJS can use breakpoints during development.

How to use?

- First setup the project as normal if you didn't have that already
- Run `pnpm playground`
- Press F5 or the play button in your `Run and debug` window
- Set breakpoints in the TresJS core code or in your playground to test in that way

### DIFF
--- a/src/composables/useSizes/index.ts
+++ b/src/composables/useSizes/index.ts
@@ -15,7 +15,19 @@ export default function useSizes(
 ) {
   const reactiveSize = toValue(windowSize)
     ? useWindowSize()
-    : useElementSize(computed(() => toValue(canvas).parentElement))
+    : useElementSize(
+      computed(
+        () => toValue(canvas).parentElement,
+      ),
+      {
+        width:
+            toValue(canvas).parentElement
+              ?.clientWidth ?? 0,
+        height:
+            toValue(canvas).parentElement
+              ?.clientHeight ?? 0,
+      },
+    )
 
   const debouncedReactiveWidth = readonly(refDebounced(reactiveSize.width, debounceMs))
   const debouncedReactiveHeight = readonly(refDebounced(reactiveSize.height, debounceMs))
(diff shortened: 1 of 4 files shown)


src/composables/useSizes/index.ts
```diff
@@ -15,7 +15,19 @@ export default function useSizes(
 ) {
   const reactiveSize = toValue(windowSize)
     ? useWindowSize()
-    : useElementSize(computed(() => toValue(canvas).parentElement))
+    : useElementSize(
+      computed(
+        () => toValue(canvas).parentElement,
```
alvarosabu, line 20: Hi @ThimoDEV why is this change needed?

alvarosabu (changes requested):

src/composables/useSizes/index.ts
ThimoDEV: There is an issue when the TresCanvas uses the parent element width and height.

The usElementSize composable from VueUse loses its reactivity I assume which causes the TresCanvas width and height to be 0. 

This issue occurs when you hit a breakpoint and continue debugging after. When you use the `useWindowSize` prop this doesn't happen.

This fixes that

ThimoDEV:
@alvarosabu Is there anything more you need/changed for this to be merged? So I can help get you there :)

src/composables/useSizes/index.ts
alvarosabu: @ThimoDEV would be possible to have this change in a separate PR, is a little bit confusing to understand the scope of this one.

alvarosabu (changes requested):

ThimoDEV:
Hi @alvarosabu , The change was needed because of HMR. When you go into a breakpoint and want to continue you expect the canvas to render what was last there. However without this change it would freeze and you would have to restart the debugger completely

alvarosabu:
Hi @ThimoDEV the are some conflicts to be resolved. I believe `useSizes` was refactored.

alvarosabu:
Due inactivity I will close this for now, if it's still relevant please feel free to reopen

PR #606, opened 2024-03-29
feat: 499 better memory management

### LINKED ISSUE #449: Memory leak caused by non disposed scene, when using routes

alvarosabu:
Hopefully closes #449 

- Recursive removal of nodes dispose (GPU) of all three disposable properties and removes refs (CPU) **if the node is not primitive**
- If it is a primitive, the user needs to handle the disposal manually by using a `dispose` function

Todo

- [x] Docs


andretchen0:
Sure thing! I'll have a look.

----

## Tests?

I feel like the answer to "Should this be disposed?" is hard to explain. And that means the implementation will be easy to break in the future.

Tests would probably help both in explaining the functionality to maintainers and in ensuring future changes don't accidentally break the functionality.

At first glance, I don't see any tests about resource disposal. Would you like me to add some tests?

alvarosabu:
Before the tests @andretchen0 , I'm struggling to dispose the whole scene on the unmount hook of the TresCanvas when changing a route, the GPU is clear but not the CPU memory allocation 🥲. I will push the latest code I used to try to make it work

andretchen0:
> Before the tests @andretchen0 , I'm struggling to dispose the whole scene on the unmount hook of the TresCanvas when changing a route, the GPU is clear but not the CPU memory allocation 🥲. I will push the latest code I used to try to make it work

Sure thing. 

Can you walk me through the steps you're taking – in the Chrome dev tools? – so that we're looking at the same data?

alvarosabu:
Hi @andretchen0 here is a video with the reproduction steps


https://github.com/Tresjs/tres/assets/4699008/bbb57f9e-6b0f-48fe-90ac-8c0b67ce9641

alvarosabu:
I think I found the reason @andretchen0, the refs kept on the CPU are entirely related to the devtools. This doesn't occur in production build.

andretchen0:
> I think I found the reason @andretchen0, the refs kept on the CPU are entirely related to the devtools. This doesn't occur in production build.

I came here to post this. Lol. 👍

alvarosabu:
Hey buddy @andretchen0 can you take a look to this one, I think I would prefer to tackle the test in the #631 task

PR #618, opened 2024-04-03
docs: correct incoherences in recipes

alvarosabu:
Some cookbook recipes had incoherences in the code samples, especially regarding the use of `Suspense` that was causing a lot of confusion across the users.


alvarosabu:
@Tresjs/docs could you please adapt the recipes in the different languages with the latest changes from the English version? I only covered the `load-models` translations

PR #632, opened 2024-04-09
chore: add pre-push linter hook

alvarosabu:
@andretchen0 😉


andretchen0 (approved):
Hey, @alvarosabu !

If you think everything's working fine over on the Cientos version, then I'm fine with adding it to Tres as well.

PR #651, opened 2024-04-17
feat: v-rotate

JaimeTorrealba:


alvarosabu:
> @alvarosabu I made this new PR (going to close the old one) changing the branch name.
> 
> But now the new linter doesn't let me pass. The problem is that it doesn't appear in my code editor, is there anything I need to do?
> 
> image image
> 
> Also the options for lint or lint:fix throw me this error
> 
> image

Which version of node are you using?

Make sure you have this extension

Screenshot 2024-04-18 at 10 12 46

Even if you editor doesnt show it, you can run the command an click on the errors in the prompt of the console and it will take you to the error, if you run `pnpm run lint:fix` should fix some of them for you. If you found yourself with troubles still call me on discord

PR #673, opened 2024-05-04
feat: 633 use loop 

### LINKED ISSUE #633: Reconsider how useRenderLoop works.

alvarosabu:
Hopefully Closes #633 

Based on the team discussion. This PR introduces a new API 

# `useLoop`

This composable allows you to execute a callback on every frame, similar to `useRenderLoop` but unique to each TresCanvas instance and with access to the context.

> [!WARNING]  
> `useLoop` can only be used inside of a `TresCanvas` since this component acts as the provider for the context data.

### Register update callbacks

The user can register update callbacks (such as animations, fbo, etc) using the `onBeforeRender`

```vue [AnimatedBox.vue]
<script setup>
import { useLoop } from '@tresjs/core'

const boxRef = ref()

const { onBeforeRender } = useLoop()

onBeforeRender(({ delta }) => {
  boxRef.value.rotation.y += delta
})
</script>
```

### Take over the render loop

The user can take over the render loop callback by using the `render` method 

```ts
const { render } = useLoop()

render(({ renderer, scene, camera }) => {
  renderer.render(scene, camera)
})
```

> [!NOTE]  
> To run this example please run the playground and go to http://localhost:5173/advanced/take-over-loop

### Register after render callbacks (ex physics calculations)

The user can also register after rendering callbacks using the `onAfterRender`

```ts
const { onAfterRender } = useLoop()

onAfterRender(({ renderer }) => {
  // Calculations
})
```

### Priority mechanism

Both `useBeforeRender` and `useAfteRender` provide an optional priority index. This indexes could be any from `[Number.NEGATIVE_INFINITY ... 0 ... Number.NEGATIVE_INFINITY]`

```ts
const { onBeforeRender, pause, resume } = useLoop()

onBeforeRender((state) => {
  if (!sphereRef.value) { return }
  console.log('updating sphere')
  sphereRef.value.position.y += Math.sin(state.elapsed) * 0.01
})

onBeforeRender(() => {
   console.log('this should happen before updating the sphere')
}, -1)
```

For example, to use Frame Object Buffer (FBO) is convenient to use a very big number to ensure that all updates happen before the FBO callback and the last one just before the scene render.

> [!NOTE]  
> To run this example (FBO) please run the playground and go to http://localhost:5173/advanced/fbo`

### Pausing and resuming the loop

The end user can use `pause` and `resume` methods:

```ts
const { onBeforeRender, pause, resume } = useLoop()

onBeforeRender(({ elapse }) => {
  sphereRef.value.position.y += Math.sin(elapsed) * 0.01
})

pause() // This stops the clock, delta = 0 and freezes elapsed.
```

Diagram onLoop

### Pausing and resuming the render

You can use `pauseRender` and `resumeRender` methods:

```ts
const { pauseRender, resumeRender } = useLoop()

onBeforeRender(({ elapse }) => {
  sphereRef.value.position.y += Math.sin(elapsed) * 0.01
})

pauseRender() // This will pause the renderer
resumeRender() // This will resume the renderer
```

## Pending

- [x] Tests
- [x] Documentation


alvarosabu:
Hi @Tresjs/core team, this PR is the result of the latest meeting and the discussion on Discord, please let me know if there is any further questions or feedback, I will wait until final agreement on the contracts to create the documentation.

I added tests as well.

andretchen0:
@alvarosabu , 

> Hi @Tresjs/core team, this PR is the result of the latest meeting and the discussion on Discord, please let me know if there is any further questions or feedback, I will wait until final agreement on the contracts to create the documentation.

You're looking for a review of code/demos? Or not yet?

alvarosabu:
Hi @andretchen0 since we had the call only the 3 of us we wanted to have your opinion about this approach, I added examples for take over mode and fbo in the advance section. 

I know also @Tinoooo wanted to discuss further the pause Render so I added the methods but didn't expose them yet.

andretchen0:
@Tinoooo 's explanation of the design is ok by me.

Glad to review whenever. Let me know.

src/core/loop.ts
```diff
@@ -0,0 +1,143 @@
...
+    isRenderPaused.value = false
+  }
+
+  function loop() {
+    const delta = clock.getDelta()
+    const elapsed = clock.getElapsedTime()
+
+    const executeCallbacks = (subscribers: Map<number, Fn[]>) => {
+      Array.from(subscribers.keys())
+        .sort((a, b) => a - b) // Ensure numerical order
```
andretchen0, line 106: For performance, it's probably best to avoid creating a new array and sorting on every tick of the loop, if possible.

src/directives/vRotate.ts
```diff
@@ -31,9 +31,8 @@ export const vRotate = {
     const quaternion = new Quaternion().setFromAxisAngle(new Vector3(x.value, y.value, z.value)
       .normalize(), radiansPerFrame)
 
-    const { onLoop } = useRenderLoop()
-
-    onLoop(() => {
+    const { onBeforeRender } = useLoop()
```
andretchen0, line 34: Directives appear to be outside of the `provide` hierarchy, so I don't think this will work directly.

In `playground/src/components/DirectiveSubComponent.vue`, try this code:

```html
<script setup lang="ts">
import { vRotate } from '@tresjs/core'
</script>

<template>
  <TresMesh v-rotate>
    <TresBoxGeometry />
    <TresMeshNormalMaterial />
  </TresMesh>
</template>
```


Error:

```
Uncaught (in promise) Error: useTresContext must be used together with useTresContextProvider
    useTresContext index.ts:296
    useLoop index.ts:31
    mounted vRotate.ts:34
```

andretchen0 (changes requested):
Hey @alvarosabu ,

This doesn't appear to implement [the API from here](https://github.com/Tresjs/tres/issues/633#issuecomment-2093237460). In particular, there doesn't seem to be a way to `off` callbacks. Is that right?

src/directives/vRotate.ts
JaimeTorrealba: @alvarosabu what do we do with this directives?

We can commented (remove from the exports) until we find a good way to handler them
Or
We can let then use `useRenderLoop`for now

alvarosabu:
> Hey @alvarosabu ,
> 
> This doesn't appear to implement [the API from here](https://github.com/Tresjs/tres/issues/633#issuecomment-2093237460). In particular, there doesn't seem to be a way to `off` callbacks. Is that right?

Let me confirm with him why he added an `off` method on that interface. We didn't discuss anything related to that haha that's why is not implemented  😅

src/composables/useLoop/index.ts
```diff
@@ -0,0 +1,64 @@
...
+    invalidate,
+    advance,
+  } = useTresContext()
+
+  function onBeforeRender(cb: Fn, index = 0) {
+    const wrappedCallback = (params: LoopCallbackParams) => {
+      cb({ ...params, camera: unref(camera) as TresCamera, scene: unref(scene), renderer: unref(renderer), raycaster: unref(raycaster), controls: unref(controls), invalidate, advance })
+    }
+    loop.register(wrappedCallback as LoopCallback, 'before', index)
+  }
```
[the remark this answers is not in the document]
alvarosabu: Actually let me confirm with Tino about that off method because it was never discussed. Not sure what he means with off.

andretchen0:
> Ok @andretchen0 I just confirmed with @Tinoooo is a method to unsubscribe the callback inspired by https://vueuse.org/shared/createEventHook/
> 
> I will need to implement it.

If it's helpful, on your branch, if you do ...

```
git fetch
git checkout fix/use-render-loop src/utils/createPriorityEventHook{.test,}.ts
```

... in `src/utils/` you'll find `createPriorityEventHook.ts`, which I'd written for the POC. 

It's like vueuse's `createEventHook`, but accepts a priority. Like `createEventHook`, it implements `off`.

It's got pretty good test coverage. I didn't write any documentation, but you can see examples in `createPriorityEventHook.test.ts`

docs/api/composables.md
```diff
@@ -4,73 +4,201 @@ Vue 3 [Composition API](https://vuejs.org/guide/extras/composition-api-faq.html#
...
+  <TresMesh ref="boxRef">
+    <TresBoxGeometry />
+    <TresMeshBasicMaterial color="teal" />
+  </TresMesh>
+</template>
+```
+
+:::
+
+Your callback function will be triggered just before a frame is rendered and it will be unmounted automatically when the component is destroyed.
```
Tinoooo, line 115:
```suggestion
Your callback function will be triggered just before a frame is rendered and it will be deregistered automatically when the component is destroyed.
```

docs/api/composables.md
```diff
@@ -4,73 +4,201 @@ Vue 3 [Composition API](https://vuejs.org/guide/extras/composition-api-faq.html#
...
+
+#### Take over the render loop
+
+The user can take over the render loop callback by using the `render` method
```
Tinoooo, line 119:
```suggestion
You can take over the render call by using the `render` method.
```

docs/api/composables.md
```diff
@@ -4,73 +4,201 @@ Vue 3 [Composition API](https://vuejs.org/guide/extras/composition-api-faq.html#
...
 
-### Pause and resume
+::: warning
+Consider that if you take over the render loop, you will need to manually render the scene and take care of features like the conditional rendering yourself.
+:::
 
-You can pause and resume the render loop using the exposed `pause` and `resume` methods.
+#### Register after render callbacks (ex physics calculations)
+
+The user can also register after rendering callbacks using the `onAfterRender`
```
Tinoooo, line 135:
```suggestion
You can also register callbacks which are invoked after rendring by using the `onAfterRender` method.
```

docs/api/composables.md
```diff
@@ -4,73 +4,201 @@ Vue 3 [Composition API](https://vuejs.org/guide/extras/composition-api-faq.html#
...
+onAfterRender(({ renderer }) => {
+  // Calculations
+})
+```
 
-// Resume the render loop
-resume()
+#### Render priority
+
+Both useBeforeRender and useAfteRender provide an optional priority index. This indexes could be any from [Number.NEGATIVE_INFINITY, Number.NEGATIVE_INFINITY] being the default index 0. The lower the index, the earlier the callback will be executed.
```
Tinoooo, line 147:
```suggestion
Both useBeforeRender and useAfteRender provide an optional priority number. This number could be anything from `Number.NEGATIVE_INFINITY` to `Number.POSITIVE_INFINITY` being the 0 by default. The lower the number, the earlier the callback will be executed.
```

docs/api/composables.md
```diff
@@ -4,73 +4,201 @@ Vue 3 [Composition API](https://vuejs.org/guide/extras/composition-api-faq.html#
...
+
+```ts
+onBeforeRender(() => {
+  console.count('before renderer')
+}, -1)
+
+onBeforeRender(() => {
+  console.count('after renderer')
```
Tinoooo, line 155:
```suggestion
  console.count('triggered first')
}, -1)

onBeforeRender(() => {
  console.count('triggered second')
```

docs/api/composables.md
```diff
@@ -4,73 +4,201 @@ Vue 3 [Composition API](https://vuejs.org/guide/extras/composition-api-faq.html#
...
+}, 1)
 ```
 
-Also you can get the active state of the render loop using the `isActive` property.
+#### Params of the callback
+
+All callbacks receive an object with the following properties:
+
+- `delta`: The delta time between the current and the last frame. This is the time in seconds since the last frame.
```
Tinoooo, line 163:
```suggestion
- `delta`: The delta time between the current and the last frame. This is the time in milliseconds since the last frame.
```

docs/api/composables.md
```diff
@@ -4,73 +4,201 @@ Vue 3 [Composition API](https://vuejs.org/guide/extras/composition-api-faq.html#
...
+#### Pausing and resuming the update loop
+
+You can use `pause` and `resume` methods:
 
 ```ts
-const { resume, isActive } = useRenderLoop()
+const { onBeforeRender, pause, resume } = useLoop()
 
-console.log(isActive) // false
+onBeforeRender(({ elapse }) => {
```
Tinoooo, line 181:
```suggestion
onBeforeRender(({ elapsed }) => {
```

docs/api/composables.md
```diff
@@ -197,46 +325,73 @@ watch(character, ({ model }) => {
 })
 ```
 
-## useTresContext
-This composable aims to provide access to the state model which contains multiple useful properties.
+## useRenderLoop
+
+The `useRenderLoop` composable can be use for animations that doesn't require access to the [context](#usetrescontext). It allows you to register a callback that will be called on native refresh rate.
```
Tinoooo, line 330:
```suggestion
The `useRenderLoop` composable can be use for animations that don't require access to the [context](#usetrescontext). It allows you to register a callback that will be called on native refresh rate.
```

docs/api/composables.md
```diff
@@ -197,46 +325,73 @@ watch(character, ({ model }) => {
...
+Be mindful of the performance implications of using this composable. It will run at every frame, so if you have a lot of logic in your callback, it might impact the performance of your app. Specially if you are updating reactive states or references.
 :::
 
-```vue
-<TresCanvas>
-  <MyModel />
-</TresCanvas>
+The `onLoop` callback receives an object with the following properties based on the [THREE clock](https://threejs.org/docs/?q=clock#api/en/core/Clock):
+
+- `delta`: The delta time between the current and the last frame. This is the time in seconds since the last frame.
```
Tinoooo, line 350:
```suggestion
- `delta`: The delta time between the current and the last frame. This is the time in milliseconds since the last frame.
```

src/composables/useLoop/index.ts
```diff
@@ -0,0 +1,64 @@
...
+    }
+    loop.register(wrappedCallback as LoopCallback, 'render')
+  }
+
+  function onAfterRender(cb: Fn, index = 0) {
+    const wrappedCallback = (params: LoopCallbackParams) => {
+      cb({ ...params, camera: unref(camera) as TresCamera, scene: unref(scene), renderer: unref(renderer), raycaster: unref(raycaster), controls: unref(controls), invalidate, advance })
+    }
+    loop.register(wrappedCallback as LoopCallback, 'after', index)
+  }
```
Tinoooo, line 52:
```suggestion
  const createWrappedCallback = (cb: Fn) => (params: LoopCallbackParams) => 
    cb({
      ...params,
      camera: unref(camera) as TresCamera,
      scene: unref(scene),
      renderer: unref(renderer),
      raycaster: unref(raycaster),
      controls: unref(controls),
      invalidate,
      advance
    })

  const onBeforeRender = (cb: Fn, index = 0) =>
    loop.register(createWrappedCallback(cb), 'before', index)

  const render = (cb: Fn) =>
    loop.register(createWrappedCallback(cb), 'render')

  const onAfterRender = (cb: Fn, index = 0) => 
    loop.register(createWrappedCallback(cb), 'after', index)
```

src/core/loop.ts
Tinoooo: I agree. Maybe an array is the better choice over a Map since it doesn't have to be sorted every time.

Tinoooo (changes requested):

src/core/loop.ts
alvarosabu: I also agree, I tried as well having the sorting indexes on an array every time you register a callback, but maybe not using Maps at all is a better option.

Also @andretchen0 provided me with another approach that he implemented on his PoC https://github.com/Tresjs/tres/pull/673#issuecomment-2096908045 with a `createPriorityEventHook`, I'm going to try that instead.

Thanks both for the  feedback 🙏🏻

alvarosabu:
> > Ok @andretchen0 I just confirmed with @Tinoooo is a method to unsubscribe the callback inspired by https://vueuse.org/shared/createEventHook/
> > I will need to implement it.
> 
> If you'd like, you can do the following on your branch:
> 
> ```
> git fetch
> git checkout fix/use-render-loop src/utils/createPriorityEventHook{.test,}.ts
> ```
> 
> Then you'll have `src/utils/createPriorityEventHook.ts`, which I'd written for the POC.
> 
> It works like `createEventHook`, including implementing `off`. It accepts a priority and sorts events according priority/insertion order.
> 
> It's got pretty good test coverage. I didn't write any documentation, but you can see examples in `createPriorityEventHook.test.ts`
> 
> If you like, you can use it in the place of the `Map`s.
> 
> Otherwise, no worries if you want to use a different implementation.

Hi @andretchen0 this was really helpful thanks a lot for this, I just pushed a refactor that uses the utility you provided instead of the subscriber's map, works like a charm so no need for `Array.from`, I also updated the tests for callback registration similar to the ones on the `createPriorityEventhook.test.ts`

I also tested a few more things, like registering the same callback twice etc and everything seems to be covered, let me know if there is anything else pending.

alvarosabu:
Question for you both @Tinoooo @andretchen0 , should I add the usage of `off` to the docs?

```
const { off, on, trigger, ...} = onBeforeRender()
```

I'm not sure we should expose `on` and `trigger` tho 🤔

Tinoooo:
> onBeforeRender

Yes, it should be in the docs.

src/composables/useLoop/index.ts
```diff
@@ -0,0 +1,64 @@
...
+    controls,
+    invalidate,
+    advance,
+  } = useTresContext()
+
+  function onBeforeRender(cb: Fn, index = 0) {
+    const wrappedCallback = (params: LoopCallback) => {
+      cb({ ...params, camera: unref(camera) as TresCamera, scene: unref(scene), renderer: unref(renderer), raycaster: unref(raycaster), controls: unref(controls), invalidate, advance })
+    }
+    return loop.register(wrappedCallback, 'before', index)
```
Tinoooo, line 37: I think there is a general flaw with the structure of how the callbacks are registered. References to the callbacks should not survive if a component that registered that particular callback is unmounted. Due to lack of time, I haven't checked, but I think this is currently the case.
@alvarosabu If you need help solving this, let me know. I think I have a solution in mind.

alvarosabu: I just tested it and the hooks seems to be unregister when the component is `unmounted`

The sphere has a v-if, the sphere has several hooks registered, when is gone the console.logs of those hooks are gone too.

Dunno if it's something that `createEventHook` does under the hood, same for @andretchen0 solution

https://github.com/Tresjs/tres/assets/4699008/c82b2661-a9c1-43a8-8dc9-1984400926b5

andretchen0: @Tinoooo @alvarosabu 

> I think there is a general flaw with the structure of how the callbacks are registered. References to the callbacks should not survive if a component that registered that particular callback is unmounted. 

Like `useEventHook`, `usePriorityEventHook` can mostly outsource this to Vue. 

Here's demo code if you'd like to verify it covers the case you're thinking about @Tinoooo . `tick` is not `console.log`ged when the component is not `v-if`d.

```html
<script setup lang="ts">
import Component from './Component.vue'
import { shallowRef } from 'vue'

const isVisible = shallowRef(true)
const animate = () => {
  isVisible.value = !!(Math.floor(Date.now() * 0.001) % 2)
  requestAnimationFrame(animate)
}

requestAnimationFrame(animate)

</script>

<template>
  <TresCanvas>
    <Component v-if="isVisible" />
  </TresCanvas>
</template>
```

```html
<script setup lang="ts">
import { useLoop } from '@tresjs/core'

useLoop().onBeforeRender(({ elapsed }) => {
  console.log('tick')
})
</script>

<template>
    <TresMesh>
      <TresBoxGeometry />
      <TresMeshNormalMaterial />
    </TresMesh>
</template>
```

----

> Hey, @Tinoooo so if I understand correctly then the end user will need to off when onBeforeUnmount right?

If we did things right, that shouldn't be necessary in regular components. It will (probably?) be necessary in things like directives if those end up using `useLoop`.

----

I should mention that I haven't looked into the guts of @alvarosabu 's implementation of `useLoop` to make sure that callbacks are removed for unmounted components – rather than simply not being called. That would be a good thing to check.

----

It would probably be best to avoid wrapping the callbacks, if possible.

andretchen0: > Dunno if it's something that createEventHook does under the hood, same for @andretchen0 solution

Here's the magic:

https://github.com/Tresjs/tres/blob/d547ab159b09eb46ad5e074c63701c3035cec12c/src/utils/createPriorityEventHook.ts#L52

src/directives/vRotate.ts
alvarosabu: @JaimeTorrealba @andretchen0 seems like if the loop is inside of the `mounted` method of the directive is called too soon (vRotate). `v-always-look-at` seems to work because the loop is inside of the method `update`. I don't think directives are meant to work with composables (setup)

Looking at the search https://github.com/search?q=v-rotate++language:vue&type=code&p=1 "nobody" at least with code on Github is using it, 99% of the search results are from another libraries aside TresJS.

If we keep them with `useRenderLoop` maybe we should add a warning in the docs that they would not work with renderMode `on-demand` or `manual`. But I will not block this feature because of the directives.

src/composables/useLoop/index.ts
alvarosabu: > I should mention that I haven't looked into the guts of @alvarosabu 's implementation of useLoop to make sure that callbacks are removed for unmounted components – rather than simply not being called. That would be a good thing to check.

@andretchen0 good news, I double-checked the same example as the video with a breakpoint on the loop to check the `count` property and when the sphere is unmounted it goes from `3` -> `1`, which is correct because the sphere registers 2 callbacks. So I can confirm it not only doesn't call them but they are also gone.


Screenshot 2024-05-07 at 19 24 06


I also checked if the sphere is mounted again after, and there are no duplicates, the count goes to 3 again which is expected

src/directives/vRotate.ts
JaimeTorrealba: unmm, no directive works well inside setup, and composables. V-rotate is the newest one, and the directive are more to "test" things, I wouldn't upload them into GitHub only in local.

Said that, I definitely don't want to block this one and/or confuse the users. If I've to take a decision, comment then (the ones who use the old loop). We can think better later.

src/composables/useLoop/index.ts
alvarosabu: @Tinoooo this is solved with the approach Andre provided

JaimeTorrealba (approved):
I don't see problems, thanks for this big work Alvaro

andretchen0:
I agree with @Tinoooo . `off` should be documented. `on` and `trigger` shouldn't be returned.

alvarosabu:
> I agree with @Tinoooo . `off` should be documented. `on` and `trigger` shouldn't be returned.

Awesome, I already made the changes on the return and document the `off` method

src/composables/useTresContextProvider/index.ts
```diff
@@ -185,6 +193,31 @@ export function useTresContextProvider({
     root: ctx,
   }
 
+  // The loop
+
+  ctx.loop.register(() => {
```
Tinoooo, line 198: **question:** Can this be moved to `useRenderer`? This would be beneficial for the seperation of concerns and the dependencies would be more obvious.

src/composables/useTresContextProvider/index.ts
```diff
@@ -176,6 +182,8 @@ export function useTresContextProvider({
     registerCamera,
     setCameraActive,
     deregisterCamera,
+    subscribers: new Map(),
```
Tinoooo, line 185: **suggestion:** I think this isn't used anywhere. It should be omitted.

src/composables/useTresContextProvider/index.ts
Tinoooo, line 198: **question:** Can/Should we make sure that this is resored once a user replaced the rendering with a custom callback and removed it afterwards via `off`?

Tinoooo (changes requested):

docs/directives/v-always-look-at.md
```diff
@@ -2,6 +2,10 @@
 
 With the new directive v-always-look-at provided by **TresJS**, you can add easily command an [Object3D](https://threejs.org/docs/index.html?q=object#api/en/core/Object3D) to always look at a specific position, this could be passed as a Vector3 or an Array.
 
+::: warning
+This directive can be only be used inside of a `TresCanvas` since this component acts as the provider for the context data and it uses `onLoop` under the hood.
```
andretchen0, line 6: Does this work? I'm getting

```
Error: useTresContext must be used together with useTresContextProvider
```

```vue
<script setup lang="ts">
import { shallowRef } from 'vue'
import { vAlwaysLookAt, vLightHelper } from '@tresjs/core'

const other = shallowRef([1,1,1])

setTimeout(() => other.value = [2,1,1], 1000)
</script>

<template>
  <TresDirectionalLight v-light-helper v-always-look-at="other" :position="[3, 3, 3]" :intensity="1" />
</template>
```

src/composables/useLoop/index.ts
```diff
@@ -0,0 +1,67 @@
...
+    loop,
+    raycaster,
+    controls,
+    invalidate,
+    advance,
+  } = useTresContext()
+
+  function wrapCallback(cb: Fn) {
+    return (params: LoopCallback) => {
+      cb({ ...params, camera: unref(camera) as TresCamera, scene: unref(scene), renderer: unref(renderer), raycaster: unref(raycaster), controls: unref(controls), invalidate, advance })
```
andretchen0, line 35: If possible, it's probably be best to avoid wrapping the callback. 

Currently, for every callback function in every frame, this code is recreating the argument object.

The argument object only needs to be created once.

src/composables/useRenderer/index.ts
```diff
@@ -274,13 +250,33 @@ export function useRenderer(
     }
   })
 
+  // Register loop
```
andretchen0, line 253: For separation of concerns, I'd move this somewhere else. 

Almost all core code has access to the renderer, so this could live just about anywhere.

On the other hand, `useRenderer` has to be passed `render`, `loop`, `invalidate`, `advance`, in order to do `loop.register`.

andretchen0 (changes requested):

src/composables/useLoop/index.ts
```diff
@@ -0,0 +1,67 @@
...
+  function onBeforeRender(cb: Fn, index = 0) {
+    const wrappedCallback = wrapCallback(cb)
+    const { off } = loop.register(wrappedCallback, 'before', index)
+    return { off }
+  }
+
+  function render(cb: Fn) {
+    const wrappedCallback = wrapCallback(cb)
+    const { off } = loop.register(wrappedCallback, 'render')
+    return { off }
```
andretchen0, line 48: This can't be `off`ed.

```playground/src/components/TakeOverLoopExperience.vue```

```html
<script setup lang="ts">
import { useLoop } from '@tresjs/core'

import { OrbitControls } from '@tresjs/cientos'
import { useControls } from '@tresjs/leches'

const { render, pauseRender, resumeRender } = useLoop()

const off = render(({ renderer, scene, camera }) => {
  console.log('on')
  renderer.render(scene, camera)
}).off

setTimeout(() => {off(); console.log('offing')}, 1000)

const { isRenderPaused } = useControls({
  isRenderPaused: {
    value: false,
    type: 'boolean',
    label: 'Pause Render',
  },
})

watchEffect(() => {
  if (isRenderPaused.value) {
    pauseRender()
  }
  else {
    resumeRender()
  }
})
</script>

<template>
  <TresPerspectiveCamera :position="[3, 3, 3]" />
  <OrbitControls />
  <AnimatedObjectUseUpdate />
  <TresAmbientLight :intensity="1" /> />
  <TresGridHelper />
</template>
```

andretchen0 (changes requested):

src/composables/useTresContextProvider/index.ts
andretchen0: Ah, I had commented the opposite.

Moving this to the `useRenderer` means:

* it has to be passed `render`, `invalidate`, `advance`, and `renderMode` which it otherwise doesn't require
* it has to know and implement the update policies, which are orthogonal to "render this scene using this camera"

If we want to keep the update in `useRenderer`, I think it'd be better off encapulated in a passed `onRender` function that `useRenderer` can simply call. But I'd personally prefer that it be removed from `useRenderer` altogether. The render policy can be part of the loop:

```
if (renderPolicy.shouldRender()) {
    renderer.render()
}
```

src/composables/useLoop/index.ts
alvarosabu: Hey @andretchen0  Check the discord discussion

docs/directives/v-always-look-at.md
alvarosabu: I'm gonna remove the directives with loop. They are not adding enough value for the troubles

src/composables/useLoop/index.ts
alvarosabu: @andretchen0 how do we get the latest context if we are returning the plain values? if we called once on the onLoop are only going to get the snapshot of the context at that moment only

src/composables/useTresContextProvider/index.ts
alvarosabu: I already moved it.... can we please agree on one way @Tinoooo @andretchen0 ?

andretchen0: @alvarosabu 

I'd prefer to reduce dependencies in `useRenderer`, but I won't block the PR because of it. It's your call.

PR #685, opened 2024-05-18
fix: manual rendering blank

alvarosabu:


andretchen0:
Somewhat related: I created a thread on the Discord about render modes. Let me know what you think.

PR #698, opened 2024-05-27
fix: cannot assign readonly properties primitives

alvarosabu:


andretchen0:
Hey @alvarosabu , it looks like your base branch didn't match the branch you're trying to merge into here maybe – 71 commits here. Just FYI.

andretchen0 (commented):
Hey @alvarosabu ,

I'm not sure I understand the issue here, so I don't know what to do to approve or request changes. 

I tried to find an open issue for more context, but didn't find one. So, I'm just going by what's changed here.

There seems to be conflicting info:

* the issue is called `fix: cannot assign readonly properties primitives`
* in `nodeOps`, the modified code only applies to arguments
* the added test only tests new arguments, but doesn't test a primitive, as far as I can tell

Fwiw, I added this test to see if read-only properties were handled, but nodeOps throws an error:

```ts
      it('handles read-only properties 2', () => {
        const obj = Object.freeze({ foo: "bar" });
        nodeOps.patchProp(obj as any, 'foo', 'bar', 'baz')
        expect(obj.foo).toBe('bar')
      })
```

... but I'm not sure that's what this is meant to fix.

PR #699, opened 2024-05-28
docs: add shared state documentation

whitespacecode:
Still need a proper thumbnail


docs/cookbook/shared-state.md
```diff
@@ -0,0 +1,113 @@
...
+author: /
+thumbnail: /recipes/animations.png
+difficulty: 0
+---
+
+# Shared State in TresJS
+
+This guide will help you get started with shared state in TresJS by building a simple scene with a cube that can be shared across component files.
+
+<StackBlitzEmbed project-id="tresjs-minimal-reproduction-rycc4j" />
```
alvarosabu, line 13: Hi @whitespacecode since the example doesn't rely on third party libs, can we replace this with a Tresjs repl playground https://play.tresjs.org/ ? 

You can find an example on how here 

https://github.com/Tresjs/tres/blob/8d53bbae37d858ddf775f0574647e86979473ac4/docs/cookbook/lights-shadows.md?plain=1#L14

docs/cookbook/shared-state.md
```diff
@@ -0,0 +1,113 @@
...
+    }
+}
+```
+
+## Setting the object to the state
+
+Next, we'll assign an object to the state and include it in a subcomponent where we can access and use it.
+
+```vue
+//Parent component
```
alvarosabu, line 40: You can take advantage of [vitepress code groups](https://vitepress.dev/guide/markdown#code-groups ) to organize several code blocks on tabs with filenames


See an example of our documentation here 

https://github.com/Tresjs/tres/blob/8d53bbae37d858ddf775f0574647e86979473ac4/docs/cookbook/load-models.md?plain=1#L68-L101

docs/cookbook/shared-state.md
```diff
@@ -0,0 +1,113 @@
...
+  <primitive v-if="mesh" :object="mesh" />
+</template>
+```
+
+## Using TresMesh components
+
+You can also add TresMesh components to the reactive state. Here, we'll use a reference and assign it to the state when mounted.
+
+```vue
+//Parent component
```
alvarosabu, line 83: Same here

alvarosabu (changes requested):

docs/cookbook/shared-state.md
whitespacecode: Hi @alvarosabu 
I tried with the playground and i'm getting this error:
`'get' on proxy: property 'modelViewMatrix' is a read-only and non-configurable data property on the proxy target but the proxy did not return its actual value (expected '#<cr>' but got '#<cr>')`
I tried a few things but i'm keep getting errors on the playground.

docs/cookbook/shared-state.md
whitespacecode: I have changed the blocks to use code groups. I also added a pinia example.

docs/cookbook/shared-state.md
whitespacecode: Ok

docs/cookbook/shared-state.md
alvarosabu: Hi @whitespacecode there is a vue tab that is empty in the latest commit

Screenshot 2024-11-05 at 15 36 38

docs/cookbook/shared-state.md
```diff
@@ -0,0 +1,154 @@
...
+import { reactive, toRefs } from "vue";
+
+const state = reactive({
+    mesh: null,
+    //you can add more objects here
+})
+
+export function useState() {
+    return {
+        ...toRefs(state)
```
andretchen0, line 37: We should slim down the reactivity here.

Feel free to point out if I'm misunderstanding, but I believe `Ref<state.mesh>` will be deeply reactive. Since a single attached THREE.Mesh contains the entire THREE.Scene graph (+ materials, geometries, and so on), that normally creates *a lot* of reactivity that goes unused.

It's fine if users want to use `Ref<Mesh>` in their own code, but we have run into concrete performance problems in the past even in relatively modest scenes. So as an official recipe, I think it would be best to refactor. 

Maybe one or both of:

* use `shallow*` alternatives
* use slimmer data, i.e. `state.isMeshVisible = true`

@alvarosabu Thoughts here?

docs/cookbook/shared-state.md
```diff
@@ -0,0 +1,154 @@
...
+      <TresMeshToonMaterial color="#4F4F4F" />
+  </TresMesh>
+
+  <SubComponent />
+</template>
+```
+
+## Using Pinia Store
+
+By using the same logic we can also use pinia.
```
andretchen0, line 110: "pinia" -> "Pinia"

["Pinia" is written with an initial cap in their docs.](https://pinia.vuejs.org/introduction.html)

andretchen0 (changes requested):
Hey @whitespacecode , thanks for the submission.

docs/cookbook/shared-state.md
alvarosabu: That's correct, better to use shallowRefs or shallowReactive. I wonder if `toRefs` automatically use `shallowRef` it when the state is `shallowReactive`  🤔

andretchen0: >  I wonder if toRefs automatically use shallowRef it when the state is shallowReactive

In the Vue docs, the type indicates that `Ref` values are returned from `toRefs`.

https://vuejs.org/api/reactivity-utilities.html#torefs

docs/cookbook/shared-state.md
whitespacecode: This should be fixed now

docs/cookbook/shared-state.md
whitespacecode: done

docs/cookbook/shared-state.md
whitespacecode: https://play.tresjs.org/#eNqVVelu2zgQfhVCwaIO1qYcNy12te4iG7dYtOiFpv+q/qClscWEIgWS8tHA794hdZhKL9cIYnOOb2Y+cmbuo48azH9VRTc1REk0N5nmlSUGbF0RweT6WRpZk0b/ppKXldKW3JNrtfsfVAlW78fkDZii+f9W6ZKJN8yC5kyQA1lpVZJHttAAj/4J/F3IBZMbZnqjK4uyWxNnSg9t3+kltwslrVbiO+YcpFUm8LiplwuFPyUqWmMah0JX6CBCbeDGYtI9OMUsUGnYUoCJjdM5h1RmShrnUWKxaP2sdx2de72T0w0TNaBOwtazMnI/AsZG5+NeN2RsdN7ApHIeN9eApOPBQlkJtMETIfOAvC2XudpODP8CJBPA9CRTQmm8sbO/Zs+vF0/8tRH8eK/3oE0FmeUbWLASNCMJVsktVxJdPv09Ju7vcxqRuHMbst+JSfPV2gwYdyZByOdcu4BKMvGarws7jDgdk9mYXLqICZcWJKr2qLigMxRlzNiJKRiW2Eaex8fiUTCPA2bwaOxeuJ+FLcU4lUuV78m98yuZXnOZkCnSS0jF8pzLdXcswCWWkIvp9A8v2PLcFv35kMozVlUN0C9sCVmy7G6tVS3z5ioScjadujgIg5fqE4zGUfP2JiWr6K1REvvOw6etAtstaQI6WdgaTpFGhbWVSeI4yyX65yD4RlMJNpZVGYfmV5d0Rmdxzo2NnRit0wipeQDcNNHvYDceCD+l04mEnaUXxyitdhjMboHdVUyeUkJv6wN0yJ2Qltz5hNhuwpyC6+yuEPEpph3/CaYM2cC5gP18Ms2B+dXFlD5BRGxH2NFySHJrhw9ZQ/4bwI3Dz6DRcJJDyU8A7Uxd8Zf0aSz4MoR0iPhCD/g0rcEpt+LrBw/TTUQuQL+rXPMOHygTQm1feZnVNfT5ZQVkd9+R35pdk/J7fCygN0j38SqxVcE26hc3b/FlBcpS5bVoL+cHyg9glKhdjo3ZNbYiph3Y+Wxf+j7DKfDRvNj5wdMWNUjU7ivw0k/t3Du+tUbw2RPn7f11uDn4I5aOlT2mjwPCH+6mU5fwqWvrV0vrZ7um0rzEab0BspnwFSbhQNy0VstbnOtHgZvPwTjGsr7JhVqDpfXpa2B+E42JVR9g1a92z2MaBfvWe2PenceopdSFToishRv2ThDHe1Xj2pAEJzwpsTFJk6fBwY1NLQ/nDhV2PoVVLf1mCvjoLksj71p2J/ehlDZpjnw2iOOk/gYPXwHXIhzH

docs/cookbook/shared-state.md
whitespacecode: Yes please feel free to change the .md file.
Most of the code is stuff i gathered from discord or by using the [stackblitz of alvarosabu](https://stackblitz.com/@alvarosabu/collections/tresjs)

alvarosabu:
Close due to inactivity.  @whitespacecode We moved the docs to a Nuxt app with a new cookbook section. Would you still be interested in creating the recipe in the new setup? I will close this PR in the meantime

whitespacecode:
Hi @alvarosabu it's almost been a year i had spend time on webgl things. I think a lot has changed in tresjs? Maybe @andretchen0 and you would beter pick this up to create a shared state page?

PR #703, opened 2024-05-31
docs: update zh/guide/getting-started.md

huzhengen:
I don't think `tree-shaking` needs to be translated into Chinese.


enpitsuLin:
Tree shaking can be translated as "除屑优化" or "摇树/摇树优化" which was common in most Chinese translations like [vue](https://cn.vuejs.org/guide/best-practices/performance.html#bundle-size-and-tree-shaking)/[rollup](https://cn.rollupjs.org/faqs/#what-is-tree-shaking), so actually here should be my mistake

Keep it untranslated also makes sense, I think maybe we need more Chinese speakers' advice, @X-smart what's your opinion?

enpitsuLin (approved):

PR #712, opened 2024-06-01
feat: add useTresReady

### LINKED ISSUE #572: Use default slot props to get Canvas context
### Description

Idea from @userquin, the idea would be to use [Vue named scoped slots ](https://vuejs.org/guide/components/slots.html#named-scoped-slots) to be able to get the canvas context on the parent like this:

```html
<TresCanvas shadows alpha v-slot="state">
  <TresPerspectiveCamera :position="[5, 5, 5]" />
  <TresOrbitControls
    v-if="state?.renderer"
    :args="[state?.camera, state?.renderer?.domElement]"
  />
</TresCanvas>
```

Or even deconstruct it

```html
<TresCanvas shadows alpha v-slot="{ camera, renderer }">
  <TresPerspectiveCamera :position="[5, 5, 5]" />
  <TresOrbitControls
    v-if="renderer"
    :args="[camera, renderer.domElement]"
  />
</TresCanvas>
```

### Suggested solution

Add this to `TresCanvas.vue`

```html
<canvas
    ref="canvas"
    :data-scene="scene.uuid"
    :class="$attrs.class"
    :data-tres="`tresjs ${pkg.version}`"
    :style="{
      display: 'block',
      width: '100%',
      height: '100%',
      position: windowSize ? 'fixed' : 'relative',
      top: 0,
      left: 0,
      pointerEvents: 'auto',
      touchAction: 'none',
      ...$attrs.style as Object,
    }"
 >
   <slot v-bind="context" />
</canvas>
```

### Alternative

_No response_

### Additional context

related #565

### Validations

### LINKED ISSUE #595: `useRenderLoop().onLoop` ticks before the renderer can meaningfully render
## Problem

`useRenderLoop().onLoop` ticks before the renderer can meaningfully render. 

When `onLoop` is initially called, e.g.:

* camera.aspect has its default value of 1, regardless of screen aspect
* renderer.domElement.width has a value of 0
* sizes.width has a value of 0

These values change on the subsequent frame, making coding/debugging more difficult than necessary.

## Solution

* `useRenderLoop()` should be dependent on the `context`'s renderer.
* The system shouldn't tick `onLoop` until the renderer is ready. 

## Context

* #251
* #252

### Reproduction

https://stackblitz.com/edit/tresjs-basic-t6zfd2?file=src%2Fcomponents%2FUseRenderer.vue

### Steps to reproduce

See StackBlitz.

### System Info

```shell
System:
    OS: Linux 5.0 undefined
    CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 18.18.0 - /usr/local/bin/node
    pnpm: 8.15.3 - /usr/local/bin/pnpm
  npmPackages:
    @tresjs/cientos: ^3.5.1 => 3.5.1 
    @tresjs/core: ^3.4.1 => 3.4.1 
    @tresjs/eslint-config-vue: ^0.2.1 => 0.2.1 
    vite: ^4.5.0 => 4.5.0
```

### Used Package Manager

npm

### Code of Conduct

andretchen0:
# Features

This PR adds the following features via a private composable `useTresReady`:

* `onTresReady` hook
* `@ready` emit from `TresCanvas`
* start loop `onTresReady`

> [!NOTE] 
> Tres is "ready" when **either**:
> * context, renderer, and canvas exist and canvas has a height and width greater than 0. 
> * 100 ms has passed since `TresContext` was created.

> [!IMPORTANT]  
> The 100 ms timeout above exists to keep Tres working as expected even in edge cases – like the user setting the canvas size to 0 x 0 explicitly.
>
> Because of the timeout, it is not guaranteed that the canvas will be renderable when `@ready`.

# Playground

```
pnpm run playground
```

... then open http://localhost:5173/basic/ready

----

# Problem

`useLoop` calls user callbacks before THREE can actually render to the screen, e.g., in the default setup, `<canvas>` will have a height/width of 0 when `useLoop` starts.

For debugging, this means that the developer needs to wait for a few frames to pass before they get "real" values in their logs/debugger.

## Solution

Start the loop and trigger `ready` callbacks when Tres is "ready".

## Use case

DX, particularly debugging. 

# Problem

#572 

Users had been asking for an easier way to get the context from a `TresCanvas` in a `<script setup>`. It currently requires a few steps that users find non-obvious, as outlined in the issue above.

## Solution

* Add a `@ready` emit that takes a callback and calls it with `TresContext` when Tres is "ready"

```vue
<script setup>
const myFn = (ctx: TresContext) => {
  // NOTE: do something with ctx
}
</script>

<template>
  <TresCanvas @ready="myFn" />
</template>
```

## Use case

DX. Advanced users who want to write a bit less code in order to access `TresContext`.

# Problem

Child components can't take advantage of "ready" without registering a loop callback.

## Solution

* Add `onTresReady` hook

```vue
<script setup>
import { onTresReady } from '@tres/core';

// NOTE: requires `TresContext`, so it can only be used in child components.
onTresReady((ctx: TresContext) => {
  // NOTE: do something with ctx
})
</script>
```

## Use case

DX. Advanced users who want to be notified when Tres is "ready" in child components.

----

## Meta

closes #595 – the present PR modifies `useLoop` not `useRenderLoop` as in the issue. But `useRenderLoop` is no longer Tres' preferred looping method
closes #572


alvarosabu (approved):
This is great @andretchen0  thank you so much. sorry took me so long to review it , I just tested everything 💚

PR #717, opened 2024-06-03
fix: Implement `createComment` and `nextSibling` nodeOps

garrlker:
Fixes #706

When a component is set to not render with `v-if`, Vue will 

1. Remove that component from the dom/scene
2. Call `createComment("v-if")`
3. Place that comment at the same spot as the previous component as a placeholder

Then when that component's `v-if` is set to true again, Vue
1. Removes that comment
2. Runs `createElement` to re-create that component
3. Places it back into it's original spot in the dom/scene

We were not implementing `v-if` so Vue did not have a placeholder comment to keep track of `v-if'd` objects positions. 
When it called `createElement` since there was no parent, we were placing it at the scene root

I'm wonder if this has not been working the entire time and none of us noticed?

PR #719, opened 2024-06-04
refactor: playground housekeeping

andretchen0:
As discussed on the Discord, this is a housekeeping PR, cleaning up playground.

## Problem

/playground/src/components is a spot for shared playground components. There are a number of components in the folder and many are only used once.

### Solution

Components only used once were reorganized into folders with their parent components

## Problem

It's useful to put an issue reproduction in the playground for local testing and for sharing. But it wasn't clear where these reproductions should go or how to signal that a component was meant to be a reproduction.

### Solution

Added an "Issues" category

### Meta

Suggested by @garrlker 

----

## Meta

This PR builds on elements added to the playground in #717 . It should be approved/merged first.


alvarosabu (approved):
Hi @andretchen0 really necessary refactor, is way cleaner now. I like the category issues, just wondering if it would get too populated eventually.

andretchen0:
> I like the category issues, just wondering if it would get too populated eventually.

Too many items in the category? I could see that happening. 

We could always create an "Issues archive" category that isn't listed on the playground landing page. But we can handle that if we get there.

alvarosabu:
> We could always create an "Issues archive" category that isn't listed on the playground landing page. But we can handle that if we get there.

Sounds good to me

PR #726, opened 2024-06-06
refactor(nodeOps): remove predicates/optional chaining, use is/context

andretchen0:
## Problem

`nodeOps` has a few cases where it does type checking that could be pushed down into the called function, making `nodeOps` simpler and the function more robust.

```ts
const deregisterCameraIfRequired = (object: TresObject) => {
  if ((object as unknown as Camera).isCamera) {
    context.deregisterCamera(object as unknown as Camera)
  }
}
```

### Solution

Push predicate into the function:

```ts
context.deregisterCamera(child)
```

## Problem

`nodeOps` contains quite a bit of optional chaining, making code more difficult to read/write than necessary.

### Solution

Factor out the need for some optional chaining, e.g., with guards.

```ts
if (!node) { return }
```

## Problem

Tres doesn't have a way to test/assert some commonly used types: `TresObject`, `Object3D`, `Fog`, etc. In some cases, it relies on THREE's `is*` fields, but this leads to more complicated/error-prone code than necessary. For example, this ...

```ts
if (child?.isObject3D) {
```

... can throw an error if the dev forgets the optional chaining (`?`) and `child` is `undefined`.

### Solution

Add `src/utils/is.ts` and accompanying tests. 

This does not require optional chaining and will not throw if `child` is `undefined`.

```ts
if (is.object3D(child)) {
```

## Problem

`nodeOps` currently does not have a `Scene` or `TresContext` by default. It must wait until a passed node arrives, containing those objects.

### Solution

Pass `TresContext` (which contains `Scene`) to `nodeOps`.


src/core/nodeOps.ts
```diff
@@ -337,10 +283,10 @@ export const nodeOps: () => RendererOptions<TresObject, TresObject | null> = ()
     patchProp,
     parentNode,
     createText: () => noop('createText'),
-    createComment,
+    createComment: () => noop('createComment'),
```
garrlker, line 286: Looks like this PR accidentally reverts #717 

Can we include those changes?

src/core/nodeOps.ts
```diff
@@ -337,10 +283,10 @@ export const nodeOps: () => RendererOptions<TresObject, TresObject | null> = ()
...
     setText: () => noop('setText'),
     setElementText: () => noop('setElementText'),
-    nextSibling,
+    nextSibling: () => noop('nextSibling'),
```
garrlker, line 289: Same as `createComment`

garrlker (changes requested):
Found a couple issues in nodeOps, looks like my recent PR changes were reverted somehow

I'm not sure I like losing types by moving checks into the functions for better clarity, but I can see what you're going for

src/composables/useCamera/index.ts
```diff
@@ -23,15 +24,21 @@ export const useCamera = ({ sizes }: Pick<TresContext, 'sizes'> & { scene: TresS
     cameras.value = [camera, ...otherCameras]
   }
 
-  const registerCamera = (newCamera: Camera, active = false) => {
-    if (cameras.value.some(({ uuid }) => uuid === newCamera.uuid)) { return }
+  const registerCamera = (maybeCamera: any, active = false) => {
```
alvarosabu, line 27: Here seems we are replacing a Type guard `newCamera: Camera` with a `isCamera(maybeCamera)`

Setting `any` is the same as not using Typescript at all, please use `unknown` type instead in all places where you don't know exactly what type to recieve

src/composables/useCamera/index.ts
```diff
@@ -23,15 +24,21 @@ export const useCamera = ({ sizes }: Pick<TresContext, 'sizes'> & { scene: TresS
...
-    if (active) { setCameraActive(newCamera) }
-    else { cameras.value.push(newCamera) }
+      if (active) { setCameraActive(camera) }
+      else { cameras.value.push(camera) }
+    }
   }
 
-  const deregisterCamera = (camera: Camera) => {
-    cameras.value = cameras.value.filter(({ uuid }) => uuid !== camera.uuid)
+  const deregisterCamera = (maybeCamera: any) => {
```
alvarosabu, line 37: Same here

src/composables/useTresContextProvider/index.ts
```diff
@@ -70,9 +70,9 @@ export interface TresContext {
    */
   advance: () => void
   // Camera
-  registerCamera: (camera: Camera) => void
+  registerCamera: (maybeCamera: any) => void
```
alvarosabu, line 73:
```suggestion
  registerCamera: (maybeCamera: unknown) => void
```

src/composables/useTresContextProvider/index.ts
```diff
@@ -70,9 +70,9 @@ export interface TresContext {
...
   setCameraActive: (cameraOrUuid: Camera | string) => void
-  deregisterCamera: (camera: Camera) => void
+  deregisterCamera: (maybeCamera: any) => void
```
alvarosabu, line 75:
```suggestion
  deregisterCamera: (maybeCamera: unknown) => void
```

src/composables/useTresEventManager/index.ts
```diff
@@ -15,8 +16,8 @@ export interface TresEventManager {
    * So we need to track them separately
    * Note: These are used in nodeOps
    */
-  registerPointerMissedObject: (object: TresObject) => void
-  deregisterPointerMissedObject: (object: TresObject) => void
+  registerPointerMissedObject: (object: any) => void
```
alvarosabu, line 19: Please don't use `any`. child coming from the nodeops is always gonna be `Object3D<Object3DEventMap>`

alvarosabu (changes requested):
Looks promising, I just checked the logic changes to lend the responsibility into the `useCamera` and `EventManager` composables. 

I still need to check the primitive changes on the nodeOps.

src/utils/is.ts
```diff
@@ -0,0 +1,40 @@
+import type { TresObject } from 'src/types'
+import type { BufferGeometry, Camera, Fog, Material, Object3D } from 'three'
+
+export function arr(a: any) {
```
alvarosabu, line 4: please replace `any` with `unknown`

alvarosabu (changes requested):

src/core/nodeOps.ts
andretchen0: For sure. Thanks!

src/composables/useCamera/index.ts
andretchen0: Sounds good. Will change.

src/composables/useTresEventManager/index.ts
andretchen0: Hey @alvarosabu , I changed `any` to `unknown` here. (Will push in a moment.)

Just to clear up this part because I don't want to merge if you're not on board with the reasoning:

> child coming from the nodeops is always gonna be Object3D<Object3DEventMap>

That was true prior to this PR.

But with this PR, the narrowing/typing is pushed down into `registerPointerMissed`.

Now `registerPointerMissed` will receive all `child: TresObject` from `insert(child: TresObject, parent: TresObject)` – not just `Object3D`. It's up to `registerPointerMissed` to decide if the passed `TresObject` requires registration. 

This is intended to make the function more robust and simplify code in the caller (`nodeOps`).

andretchen0:
Hey @alvarosabu , @garrlker 

Thanks for the reviews! I think all the raised issues have been covered with the latest push.

garrlker (approved):
Looks good :)

alvarosabu (approved):
Thanks for fixing the comments, I still think we can improve the typing with the correct types but for now `unknown` is ok.

Issue #728, opened 2024-06-07
`Event` is not triggered in `TresGroup`


hexianWeb:
### Describe the bug

I tried writing this code into `Tresjs v 4.0.2` I think this issue and issue #727 have a common cause
```vue
    <TresGroup>
      <TresMesh
        :position="[0, 0, 0]"
        @click="onClick"
        @double-click="onDoubleClick"
        @pointer-enter="onPointerEnter"
        @pointer-leave="onPointerLeave"
        @pointer-move="onPointerMove"
        @context-menu="onContextMenu"
        @pointer-missed="onPointerMissed"
      >
        <TresBoxGeometry :args="[1, 1, 1]" />
        <TresMeshToonMaterial color="#efefef" />
      </TresMesh>
    </TresGroup>
```
Only `@pointer-missed` has been triggered during this process

https://github.com/Tresjs/tres/assets/59913119/8ca05964-6e9d-4443-a1d8-7e543aebb5fe

### Reproduction

https://stackblitz.com/~/github.com/hexianWeb/starter

### Steps to reproduce

switch the branch to the `bug` branch, or just copy this code into `TheExperience.vue`
```vue 
    <TresGroup>
      <TresMesh
        :position="[0, 0, 0]"
        @click="onClick"
        @double-click="onDoubleClick"
        @pointer-enter="onPointerEnter"
        @pointer-leave="onPointerLeave"
        @pointer-move="onPointerMove"
        @context-menu="onContextMenu"
        @pointer-missed="onPointerMissed"
      >
        <TresBoxGeometry :args="[1, 1, 1]" />
        <TresMeshToonMaterial color="#efefef" />
      </TresMesh>
    </TresGroup>
```
and remove the `TresGroup` near camera
```vue
  <TresGroup>
    <TresPerspectiveCamera :position="[5, 5, 5]" />
  </TresGroup>
```

### Used Package Manager

pnpm

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/Tresjs/tres/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/Tresjs/tres/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://tresjs.org/guide).
- [X] Check that there isn't [already an issue](https://github.com/tresjs/tres/issues) that reports the same bug to avoid creating a duplicate.
[...]

PR #731, opened 2024-06-08
fix: group should recursive search for child elements (#728)

hexianWeb:
**original func**
Mesh with events cannot be collected correctly when they wrapped by `TresGroup`, So when collections meet `Tresgroup`,  should recurse  to the child elements.


src/composables/useRaycaster/index.ts
```diff
@@ -16,7 +16,16 @@ export const useRaycaster = (
...
+  const objectWihtEvents = computed(() => {
+    const hasEvents = (object)=> object.__tres?.eventCount > 0;
+    const checkObject = (object)=> {
+      if (object.isGroup) {
+        return object.children.some(checkObject)
+      }
+      return hasEvents(object)
+    }
+    return objects.value.filter(checkObject)
+  })
```
garrlker, line 28: Hey @hexianWeb, thank you for the issue and the PR with a fix!

The solution looks good as is, but I'd like to clean it up a bit for readability

I know it's subjective on whether ternaries *are* more readable, so feel free to push back if you don't like this suggestion

```suggestion
  const hasEvents = (object) => object.__tres?.eventCount > 0;
  const isGroup = (object) => object.isGroup ? object.children.some(isGroup) : hasEvents(object)

  // Recurse through scene graph and filter for objects with events
  const objectWihtEvents = computed(() => objects.value.filter(isGroup))
```

Also, while this fixes the `Groups` case, any `Object3D` can contain children as well, so maybe the fix will look more like this?

```suggestion
  const hasEvents = (object) => object.__tres?.eventCount > 0;
  const hasChildren = (object) => object?.children.length > 0 ? object.children.some(hasChildren) : hasEvents(object)

  // Recurse through scene graph and filter for objects with events
  const objectWihtEvents = computed(() => objects.value.filter(hasChildren))
```

And 2 things that you didn't cause but would be nice if you could fix

1. Can we fix the `objectWihtEvents` typo? `objectWithEvents`
2. We're filtering for the events in the raycaster, but the source of these objects is coming from the EventManager. Can we move this logic into it's computed instead? [Link](https://github.com/Tresjs/tres/blob/main/src/composables/useTresEventManager/index.ts#L33)

garrlker (changes requested):
Overall looks good! A couple suggestions and questions

src/composables/useRaycaster/index.ts
hexianWeb: Sorry for replying so late, I have changed the code. 
Please let me know if any code needs to be modified.

garrlker (approved):

hexianWeb:
Merging is blocked! Is there anything I should do now?

src/composables/useRaycaster/index.ts
```diff
@@ -9,14 +9,13 @@ import type { DomEvent, TresCamera, TresEvent } from 'src/types'
 import type { TresContext } from '../useTresContextProvider'
 
 export const useRaycaster = (
-  objects: Ref<Object3D[]>,
+  objectWithEvents: Ref<Object3D[]>,
```
andretchen0, line 12: Can we please change this to `objectsWithEvents`? (The original was plural and the type is an array, so many `objects`.)

Thanks! And sorry for the late change request here.

andretchen0 (changes requested):

src/composables/useRaycaster/index.ts
hexianWeb: Thank you for your prompt response

Issue #693, opened 2024-05-24
Add userData to vue devtools component inspector


alvarosabu:
### Description

As a developer I would like to inspect the userData object when I select an object to inspect on the devtools,

### Suggested solution

Just add the property

### Validations

- [X] I agree to follow this project's [Code of Conduct](https://github.com/Tresjs/tres/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/Tresjs/tres/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://tresjs.org/guide).
- [X] Check that there isn't [already an issue](https://github.com/tresjs/tres/issues) that reports the same bug to avoid creating a duplicate.

PR #740, opened 2024-06-20
feat: (devtools) add userData to inspectable properties

DennisSmuda:
This PR should take care of [this issue](https://github.com/Tresjs/tres/issues/693) if I understood it correctly 🤞 

Adds the `userData` object so it can be inspected via devtools:

Screenshot 2024-06-20 at 22 21 53


src/devtools/plugin.ts
```diff
@@ -249,7 +249,11 @@ export function registerTresDevtools(app: DevtoolsApp, tres: TresContext) {
                 key: 'matrixWorld',
                 value: instance.matrixWorld,
               },
-
+              {
+                key: 'userData',
```
alvarosabu, line 253: Maybe we can enhance this further and loop on the key-value pairs of the instance instead of going one by one like this:

```ts
payload.state = {
  object: Object.entries(instance).map(([key, value]) => {
    return {
      key,
      value,
      editable: true,
    }
  }).filter(({ key }) => key !== 'parent'),
}
```

alvarosabu (changes requested):

src/devtools/plugin.ts
DennisSmuda: @alvarosabu absolutely, thanks for the feedback, I'll get on it.

DennisSmuda: Wow I didn't even have to do anything :)

Screenshot 2024-07-07 at 22 52 35

alvarosabu (approved):
@DennisSmuda awesome job! Thanks 🙏🏻

PR #747, opened 2024-06-27
docs: align docs and donut example code

m1911star:
since https://github.com/m1911star/tres/blob/main/docs/.vitepress/theme/components/DonutExample.vue here can rotate with orbitcontrols, just align the code.

also, I just wonder where to modify the codes in related playground


alvarosabu:
Hi, @m1911star is actually a documentation design decision, we didn't want to introduce such a complex topic in the first scene. We will not proceed with the change

Issue #701, opened 2024-05-30
`<primitive> :object` breaking references to THREE objects


andretchen0:
## Problem

When using a THREE object in `<primitive :object="[the THREE object]">`, the on-screen object cannot be updated in `<script setup>`.

## Possible cause

It appears that `src/core/nodeOps.ts` clones the `primitive`'s `:object`, breaking the reference to the THREE object in `<script setup>`.

### Reproduction

https://stackblitz.com/edit/tresjs-basic-lgdhxs?file=src%2Fcomponents%2FTheExperience.vue

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/Tresjs/tres/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/Tresjs/tres/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://tresjs.org/guide).
- [X] Check that there isn't [already an issue](https://github.com/tresjs/tres/issues) that reports the same bug to avoid creating a duplicate.
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

andretchen0:
Related:

* https://github.com/Tresjs/tres/issues/363
* https://github.com/Tresjs/tres/issues/503

PR #749, opened 2024-06-29
fix: attach detach

### LINKED ISSUE #496: Allow `TresObject3D#material` to set Array
### Description

Currently, `TresObject3D#material` doesn't allow to set array.  My suggestion is to make Array acceptable as well.

https://github.com/Tresjs/tres/blob/main/src/types/index.ts#L43

```
- material?: THREE.Material & TresBaseObject
+ material?: THREE.Material & TresBaseObject | (THREE.Material & TresBaseObject)[]
```

## (Personal) reason

I use TresJS and [spine-ts THREE.JS](https://github.com/EsotericSoftware/spine-runtimes/blob/4.1/spine-ts/spine-threejs/README.md). 

THREE.Mesh#material's type is `Material | Material[]`. (https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/three/src/objects/Mesh.d.ts#L24)

spine-ts' set Array material according to THREE's type. (https://github.com/EsotericSoftware/spine-runtimes/blob/4.1/spine-ts/spine-threejs/src/MeshBatcher.ts#L60)

However, `TresObject3D#material` doesn't allow to set Array material, when a spine's mesh is rendered in Scene of TresJS and a material**s** in spine's mesh are evaluate, TresJS throws runtime error.

## Basis that my suggestion is reasonable

A type of THREE.Mesh#material extending THREE.Object3D is `Material | Material[]`, so I think it's reasonable that a type of `TresObject3D#material` extending THREE.Object3D is `THREE.Material & TresBaseObject | (THREE.Material & TresBaseObject)[]`.

### Suggested solution

https://github.com/Tresjs/tres/blob/main/src/types/index.ts#L43

```
- material?: THREE.Material & TresBaseObject
+ material?: THREE.Material & TresBaseObject | (THREE.Material & TresBaseObject)[]
```

and adding logic for array.

### Alternative

nothing. I currently use pnpm patch and add logic for array.

### Additional context

I can create PR for this suggestion.

### Validations

andretchen0:
closes #496 

## Context

This PR deals with "attached objects", e.g., materials, geometries. 

These objects can't be `add()`ed via THREE's `Object3D.add(value)`. Instead, they are "attached" as values as in `someObject[someField] = value`. 

Compared to `add()`ing, "attach"ing requires extra implementation steps and logic in order to support Vue's declarative paradigm transparently for users.

This PR is another in a series to prepare for a rework of Tres' `<primitive />`. Issue: #701 

## Playground

http://localhost:5173/advanced/material-array

## Problem

Fields with attached objects don't revert back their original values when unmounted. As a result, an unmounted material will continue to affect its parent.

### Solution

When attaching, save the previous value to a variable and restore it when detaching.

## Problem

The current implementation of `:attach` doesn't allow for array indices. This means e.g., it's not possible to attach multiple materials, even though THREE allows this. #496

### Solution

Add "attach" logic to traverse arrays.

## Problem

`:attach` is a prop but isn't reactive, as users might expect.

### Solution

Implement reactivity for `:attach`.

## Problem

`:attach` prop value can't be a "pierced prop", i.e., it can only replace values directly on the parent – `parent[field]`. It  can't be used to "drill down" into `parent[object][field]`.

### Solution

Implement "pierced" logic for `:attach` – see `resolve` in `src/utils/index.ts` and tests in `src/utils/index.test.ts`.

## Refactor

THREE does not define the attach field. Tres defines it. Most Tres data is stored in `object.__tres`. `attach` is stored directly on `object` as `object.attach`.

Refactor and store `:attach` value in `object.__tres.attach`.

## Refactor

Move `src/utils/nodeOpsUtils.ts` to `src/utils/index.ts`.


alvarosabu (approved):
Amazing work @andretchen0 thanks for solving this 🙏🏻

garrlker (approved):
Sorry, it took me a while to review the full code changes
I love the attention to detail in your unit tests. No issues here

I also ran through the playground suites as well for more of a hands on verification and everything looks good.

Approved!

Issue #750, opened 2024-06-29
Deltatime Documentation Inaccuracy


nart4hire:
### Describe the bug

In the docs, delta is described as milliseconds from last frame, which is inaccurate since under the hood it uses Clock.deltaTime() which is in seconds.

### Reproduction

https://stackblitz.com/edit/tresjs-minimal-reproduction-tuauqe?file=src%2Fcomponents%2FTheExperience.vue

### Steps to reproduce

Just use the useLoop function as described in the documentation

### System Info

```shell
System:
    OS: Windows 11 10.0.22631
    CPU: (20) x64 12th Gen Intel(R) Core(TM) i7-12700K
    Memory: 21.41 GB / 31.77 GB
  Binaries:
    Node: 21.7.1 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.19 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 10.5.0 - C:\Program Files\nodejs\npm.CMD
    pnpm: 8.6.2 - C:\ProgramData\chocolatey\bin\pnpm.EXE
  Browsers:
    Edge: Chromium (126.0.2592.81)
    Internet Explorer: 11.0.22621.3527
```

### Used Package Manager

pnpm

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/Tresjs/tres/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/Tresjs/tres/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://tresjs.org/guide).
- [X] Check that there isn't [already an issue](https://github.com/tresjs/tres/issues) that reports the same bug to avoid creating a duplicate.
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #752, opened 2024-06-30
chore(docs): onLoop delta definition fix

nart4hire:
Fixes Issue #750 

Double checked the other languages, but they all seem to be correct so only the English version needed change.


alvarosabu (approved):
You are absolutely right @nart4hire thanks for correcting it

alvarosabu (approved):

PR #757, opened 2024-07-03
feat(app): 680 UseTexture composable as component

### LINKED ISSUE #680: UseTexture as component
### Description

As a developer, what if we provide a function similar to [vuesuse](https://vueuse.org/core/useMouse/#component-usage)

In which we can use a composable as a component too. This could be maybe handy for others composable. 

### Suggested solution

EJ...
```html
<Suspense>
   <UseTexture :map="MYMAPTEXTURE" v-slot="{ map }" />
      <TresMeshStandardMaterial :map="map"
   <UseTexture />
</Suspense>
```
### Alternative

Could be similar to: https://github.com/vueuse/vueuse/blob/main/packages/core/useMouse/component.ts

Any other alternative is considered

### Additional context

What do you think?

### Validations

JaimeTorrealba:
closes #680 

Convert a composable like this one, as a component is not difficult, what I would like to know is what do you think about this one

## Component Name

Vueuse, use the same name of the composable and seems to have no problem (the component start with uppercase), but what do you think? Could this confuse users? Any other alternatives.

useTexture and UseTexture

```html
    <Suspense>
      <UseTexture v-slot="{ map }" :map="path">
        <TresMesh>
          <TresBoxGeometry />
          <TresMeshStandardMaterial :map="map" />
        </TresMesh>
      </UseTexture>
    </Suspense>
```

## Props and result

 They share the same name in the example they are :map="map" and the result currently is v-slot="{ map}"

I personally don't see any problem but again I would like to know what do you think?

## Docs

I would not do the docs just yet, until your answers. But my idea is to put the docs of this just below the existing composable, any thoughts?


src/composables/useTexture/component.ts
alvarosabu, line 1: Why not follow the same structure as `useFBO`, `AnimatedSprite`, `useSurfaceSmapler` and the rest?
Screenshot 2024-08-21 at 13 00 00

We need to keep the code consistent

Please refactor the component not to use `defineComponent` and use a `component.vue` setup syntax approach

alvarosabu (changes requested):
Love the idea but not the execution. We need to be more strict on the standardization of the code base and avoid having 3 different approaches to components.

alvarosabu (approved):

PR #764, opened 2024-07-08
fix(primitive): implement as proxy to avoid breaking references 

### LINKED ISSUE #701: `<primitive> :object` breaking references to THREE objects

### LINKED ISSUE #721: Cannot read properties of undefined (reading 'dispose') - GridHelper
### Describe the bug

When i try to show/hide the GridHelper i get a TypeError. The code still works. 
I tried it on an older tresjs version and no typeError happend.

`<TresGridHelper v-if="gridHelper.visible" :args="[10, 100, 0xff0000]" />`
image

Not sure if that is also the case for other components.

### Reproduction

https://stackblitz.com/edit/tresjs-basic-hfgtuj

### Steps to reproduce

Use the reproduction, click on the button 'Toggle Grid' and check the console logs.

### System Info

```shell
System:
    OS: macOS 14.3.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 90.73 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.18.0 - ~/.nvm/versions/node/v18.18.0/bin/node
    Yarn: 1.22.21 - ~/.nvm/versions/node/v18.18.0/bin/yarn
    npm: 10.6.0 - ~/Code/Modulab/node_modules/.bin/npm
  Browsers:
    Chrome: 125.0.6422.142
  npmPackages:
    @tresjs/cientos: ^3.9 => 3.9.0
    @tresjs/core: ^4.0.2 => 4.0.2
    vite: ^5.2.10 => 5.2.10
```

### Used Package Manager

npm

### Code of Conduct

### LINKED ISSUE #751: Rigged models are squished and broken
### Describe the bug

Adding a rigged model with `useGLTF` looks squished and broken

Screenshot 2024-06-29 at 11 53 14

### Reproduction

https://stackblitz.com/~/edit/tresjs-squished-models?file=src/App.vue

### Steps to reproduce

_No response_

### System Info

```shell
System:
    OS: Linux 5.0 undefined
    CPU: (4) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 18.20.3 - /usr/local/bin/node
    pnpm: 8.15.6 - /usr/local/bin/pnpm
  npmPackages:
    @tresjs/cientos: 3.9.0 => 3.9.0 
    @tresjs/core: 4.0.2 => 4.0.0 
    vite: ^5.3.2 => 5.3.2
```

### Used Package Manager

npm

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/Tresjs/cientos/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/Tresjs/cientos/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://cientos.tresjs.org/guide).
- [X] Check that there isn't [already an issue](https://github.com/Tresjs/cientos/issues) that reports the same bug to avoid creating a duplicate.

andretchen0:
This is the last in a series of PRs to implement primitives in a way that doesn't break references in `<script setup>` (see #701). With this PR, `<primitive>` is now implemented as a JS `Proxy`.

## Problem

#701 Under the hood, `<primitive />` copies its `:object`. This breaks references to the object in `<script setup>`.

### Solution

Implement `<primitive />` as a JS `Proxy`.

### Playground

http://localhost:5173/issues/701

## Problem

#751 Rigged models are squished and do not animate. Bones and animations lose their reference to the model.

### Solution

Implement `<primitive />` as a JS `Proxy`.

### Playground

http://localhost:5173/models/rigged

## Problem

#721 Tres typically removes/disposes in descendants-first order – children are removed before parents. `nodeOps.remove` sometimes calls a THREE object's `dispose` method. THREE's `dispose` implementation does not check for the existence of attributes before trying to delete them – which can result in a user-facing error. 

### Solution

Try/catch on `dispose` in `nodeOps.remove`.

## Problem

Cientos' components using primitives under the hood – Sparkles and Contact Shadows – do not work under Tres `v4`.

### Solution 

Implement `<primitive />` as a JS `Proxy`.

### Playground

http://localhost:5173/issues/701-cientos-v4

closes #751
closes #721
closes #701


src/composables/useTresContextProvider/index.ts
```diff
@@ -182,6 +183,7 @@ export function useTresContextProvider({
     setCameraActive,
     deregisterCamera,
     loop: createRenderLoop(),
+    isTresContext: true,
```
alvarosabu, line 186: Hi @andretchen0 what's the purpose of the `isTresContext`? Couldn't find any reference of usage on your changes

src/core/nodeOps.ts
```diff
@@ -68,139 +87,145 @@ export const nodeOps: (context: TresContext) => RendererOptions<TresObject, Tres
...
-    if (!isPrimitive) {
-      // NOTE: In recursive `remove`, the array elements will
-      // remove themselves from these arrays, resulting in
-      // skipped elements. Make shallow copies of the arrays.
-      if (node.children) {
-        [...node.children].forEach(child => remove(child, shouldDispose))
+    // NOTE: 4) Dispose `node`
+    if (shouldDispose && !is.scene(node)) {
+      if (is.fun(dispose)) {
+        dispose(node)
```
alvarosabu, line 207:
```suggestion
        dispose(node as TresInstance)
```

src/core/nodeOps.ts
```diff
@@ -68,139 +87,145 @@ export const nodeOps: (context: TresContext) => RendererOptions<TresObject, Tres
...
+          node.dispose()
+        }
+        catch (e) {
+          // NOTE: We must try/catch here. We want to remove/dispose
+          // Vue/THREE children in bottom-up order. But THREE objects
+          // will e.g., call `this.material.dispose` without checking
+          // if the material exists, leading to an error.
+          // See issue #721:
+          // https://github.com/Tresjs/tres/issues/721
+          // Cannot read properties of undefined (reading 'dispose') - GridHelper
```
alvarosabu, line 220: Should we add a warning here at least? @andretchen0

playground/src/pages/advanced/disposal/index.vue
alvarosabu, line 1: Hi, @andretchen0 this route was missing from the router's file, so it was not available in the UI of the playground, so I added it

I did run it but not sure if it's supposed to have a checkbox in order to see the v-if change.
Screenshot 2024-07-09 at 15 09 49

alvarosabu (changes requested):
Hi @andretchen0 I think I'm gonna pin this PR somewhere just as an example of how a PR should be done, thanks for the incredible amount of work, this is how primitives should have been working fro the beginning but now thanks to you they work.

src/composables/useTresContextProvider/index.ts
andretchen0: Hey @alvarosabu , good catch.

I had started to tackle general-purpose disposal for primitives.

Tres objects have a link to `TresContext` in their `__tres`. If we recurse through it to dispose, we'll end up disposing `Scene` and `Renderer`. So I added `is.tresContext` so we have an easy predicate to check against.

In the end, I decided against putting general purpose disposal in this PR, since it's a standalone function. But I left the `is.tresContext` and the field here.

Would you like me to remove it?

playground/src/pages/advanced/disposal/index.vue
andretchen0: That's strange. Is the link showing up for you?

Screenshot 2024-07-09 at 17 25 10 copy

The route should be: http://localhost:5173/issues/717vIf

At least in my local version of playground/src/router/routes/issues.ts , the route is the third in this array:

```js
export const issuesRoutes = [
  {
    path: '/issues/701',
    name: '#701: primitive :object',
    component: () => import('../../pages/issues/701/index.vue'),
  },
  {
    path: '/issues/701-cientos-v4',
    name: '#701: <primitive> in Cientos v4',
    component: () => import('../../pages/issues/701-cientos-v4/index.vue'),
  },
  {
    path: '/issues/717vIf',
    name: '#717: v-if',
    component: () => import('../../pages/issues/717/index.vue'),
  },
  {
    path: '/issues/749-attach-detach',
    name: '#749: attach-detach',
    component: () => import('../../pages/issues/749/index.vue'),
  },
]

```

----

There's no checkbox. On its own, the box should flash on and off, like this:

https://github.com/Tresjs/tres/assets/20469369/8a90749d-29cc-4a26-9182-f31214fe87f8

I've found it slightly simpler *not* to require user interaction for testing issues, because that means I don't have to interact with the window in order to see if things are working/broken.

src/core/nodeOps.ts
andretchen0: Looks good. Pushed the change.

src/composables/useTresContextProvider/index.ts
alvarosabu: Yep let's remove it 👍

andretchen0: Hey @alvarosabu 

Ok, I'll go ahead and remove it here.

I'm working on disposal right now and will need to *not* recurse through a `TresContext`. Is it ok with you if I put the field back in that PR?

We *can* duck type a `TresContext` from its existing fields, but I'd prefer to have `isTresContext` since ...

* an `is[...]: true` field is how most THREE objects identify themselves
* duck typing on the existing fields of `TresContext` leaves us open to breaking `is.tresContext` with a refactor that changes its fields. We're unlikely to change `isTresContext: true`, because this is the *only* purpose it serves.

playground/src/pages/advanced/disposal/index.vue
alvarosabu: Ohhh then my confusion @andretchen0  because there is another route inside `/advanced/disposal` which is similar but wasn't working as you show in the video. All good

src/composables/useTresContextProvider/index.ts
alvarosabu: Deal

alvarosabu (approved):

PR #768, opened 2024-07-11
feat(TresCanvas): add dpr prop

### LINKED ISSUE #765: Library override the DPR setting in autonomy without chances to clamp / control it
I noticed that Tres set the DPR ([see code](https://github.com/Tresjs/tres/blob/00bef337f189b04ee3b77a4ce0afea63ce2973b2/src/composables/useRenderer/index.ts#L152C26-L152C45)) completely in autonomy without any way to  control that, imho is conceptually wrong for couple of reasons:

1. The dev should be in charge to decide the DPR or **at least be able to clamp**. For example if the app is opened on a 3DPR mobile is gonna run at 3, that is completely insane.
2. The app should not change the dpr on resize without consent. I noticed it because i was setting manually in a range 
 ```js 
   const dpr = MathUtils.clamp(window.devicePixelRatio, 1, 1.4) 
  renderer.value.setPixelRatio(dpr)
  ```
  but on resize switching between monitors on my macbook was set back to 2 due the `useDevicePixelRatio` hook. This second point tho depends on the first one. If We decide to let the app handle the resize, **it's 100% needed** to add a `clamp` on that.
  
This can lead to some big issues, both perf wise and effect wise. I notice the "issue" because i was doing some effect in screen-space, where my resolution wasn't matching my DPR `vec2 st = gl_FragCoord.xy / (gResolution * gDpr);` 
  
  My take on this is that the library can take care of the DPR but at least give the user some way to control, like they don in `r3f` where you can set `dpr=[low, high]` that clamps it :)
  
  

### Reproduction

---

### Steps to reproduce

----

### System Info

```shell
----
```

### Used Package Manager

npm

### Code of Conduct

andretchen0:
## Problem

Tres sets the renderer's `pixelRatio` automatically, but doesn't allow the developer to specify or influence the value.

## Solution

Add a `:dpr` prop to `<TresCanvas />` that:

* can be omitted, defaulting to current behavior.
* can be `number`: overriding the system's `pixelRatio`.
* can be `[number, number]`, clamping the system's `pixelRatio` between the first (expected min) and second (expected max) values.

## Playground

Playground demo added at:

http://localhost:5173/advanced/device-pixel-ratio

## Meta

closes #765


alvarosabu (approved):

alvarosabu:
You are awesome @andretchen0

dghez:
just FYI, is missing on the docs https://docs.tresjs.org/api/tres-canvas.html

dghez:
Left a new comment on the main thread https://github.com/Tresjs/tres/issues/765 so i easier to track

PR #772, opened 2024-07-12
feat: deprecate disable render

alvarosabu:


Tinoooo:
@alvarosabu The prop doesn't work at all for me now. Should we remove it completely? Marking it deprecated would imply it still works.

Tinoooo:
The prop will be removed in v5.

PR #787, opened 2024-07-20
Add option to use `WebGPURenderer`

kekkorider:
This feature adds a `webGPU` parameter to `TresCanvas` which determines what renderer to use, whether the `WebGLRenderer` or `WebGPURenderer`, allowing people to use new techs like TSL.

At the moment this feature is broken as soon as I import the `WebGPURenderer`, it returns the following error:

```bash
✘ [ERROR] Top-level await is not available in the configured target environment ("chrome87", "edge88", "es2020", "firefox78", "safari14" + 2 overrides)

    ../node_modules/.pnpm/three@0.166.1/node_modules/three/examples/jsm/capabilities/WebGPU.js:14:15:
      14 │   isAvailable = await navigator.gpu.requestAdapter();
```

The error can be seen by uncommenting [this line](https://github.com/kekkorider/tres-webgpu-renderer/blob/feat/webgpu-renderer/src/composables/useRenderer/index.ts#L2).

I tried fixing it with the [`vite-plugin-top-level-await`](https://www.npmjs.com/package/vite-plugin-top-level-await) plugin bun unfortunately no luck,

I assume there's some configuration somewhere I'm not aware of that I should override.

### DIFF
--- a/package.json
+++ b/package.json
@@ -107,6 +107,7 @@
     "vite-plugin-dts": "3.9.1",
     "vite-plugin-inspect": "^0.8.4",
     "vite-plugin-require-transform": "^1.0.21",
+    "vite-plugin-top-level-await": "^1.4.2",
     "vite-svg-loader": "^5.1.0",
     "vitepress": "1.3.0",
     "vitest": "^2.0.2",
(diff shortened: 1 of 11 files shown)


alvarosabu:
cc @andretchen0

PR #795, opened 2024-08-02
fix: onPointerEnter event, objects can be use compare the newly added execution of enter

hero8080:
fix: onPointerEnter event, objects can be use compare the newly added execution of enter


alvarosabu:
Hi @hero8080, on v5 we are leveraging the pointer event system with `@pmndrs/pointer-events` package. So this fix is no longer needed

PR #806, opened 2024-08-27
feat: pass all provides down to custom renderer

### LINKED ISSUE #537: Unable to use custom provide inject
### Describe the bug

Custom injection cannot be used in components
After checking the source code, it seems that except for the TresCanvas component, other components cannot be seen in devTools, so there is no concept of sub-components.
Need to extend the context with a custom injection?

### Reproduction

https://stackblitz.com/edit/vitejs-vite-8p7urb?file=src%2Fcomponents%2FCustomizeContext.vue

### Steps to reproduce

_No response_

### System Info

```shell
system:
    OS: windows
    CPU: (8) x64 Intel(R) Core(TM) i7-8700 CPU @ 3.2GHz
    Memory: 0 Bytes / 0 Bytes
  Binaries:
    Node: 18.18.0
    Yarn: 1.22.19
    npm: 9.4.2
    pnpm: 8.14.0
  npmPackages:
    @tresjs/cientos: 3.7.0 => 3.7.0 
    @tresjs/core: 3.7.0 => 3.7.0 
    vite: ^5.0.12 => 5.0.12
```

### Used Package Manager

npm

### Code of Conduct

### LINKED ISSUE #732: Impossible to `inject` variable in canvas child component
### Describe the bug

I'm trying to `inject` a variable inside a component in a `TresCanvas`, but it's always `undefined`.

In the example i log the variable provided by the first component but in the `SecondTresjs` component, the variable is `undefined`

Am i doing something wrong? I cannot find anything in the docs about the context.

### Reproduction

https://stackblitz.com/edit/vitejs-vite-gtrfxh?file=src%2Fcomponents%2FFirstTresjs.vue

### Steps to reproduce

_No response_

### System Info

_No response_

### Used Package Manager

npm

### Code of Conduct

alvarosabu:
Closes #537 
Closes #732 

This PR adds an opt-out bridge to pass down Vue parent `provides` down to the Tres components via `TresCanvas` with a flag to disable it `enableProvideBridge`

## How to test

- `pnpm run dev`
- Go to [#732: provide/inject](http://localhost:5173/issues/732)
- Open devtools
- Click the button

You should see console logs with prefixes like `tres` from the child tres component with the injects and `vue` for the vue child components, both should show a value different from undefined. 

If you pass the `:enable-provide-bridge="false"` and refresh, the `tres` consoles should be undefined.


src/components/TresCanvas.vue
```diff
@@ -107,17 +107,42 @@ const canvas = ref<HTMLCanvasElement>()
...
 const createInternalComponent = (context: TresContext, empty = false) =>
   defineComponent({
     setup() {
       const ctx = getCurrentInstance()?.appContext
-      if (ctx) { ctx.app = instance as App }
+      if (ctx) { ctx.app = instance?.appContext.app as App }
       provide('useTres', context)
       provide('extend', extend)
 
+      const provides = {}
```
andretchen0, line 121: I think this is an improvement over the current situation where ancestors can't `provide` to `TresCanvas` components.

But it's got some limitations:

----

Because the solution here copies values, updates in ancestor `provide`s won't be reflected here unless:

* the provided value is non-primitive
* the provided value is updated via its keys

----

`provide`s in ancestors won't be `provide`d if they don't exist at the time of the instantiation of the `TresCanvas`.

----

Not sure if either of those are problems we want to address. But they should probably be documented if we don't address them.

andretchen0 (changes requested):

playground/.eslintrc-auto-import.json
andretchen0, line 1: Was this file `git add`ed by mistake maybe?

andretchen0 (changes requested):
Hey @alvarosabu ,

I grabbed the latest revision. `provide` from outside the `<TresCanvas>` doesn't seem to work on components within the `<TresCanvas>` – I tried adding an `inject` in `SubComponentWithInject.vue`.

I could be wrong though. I didn't find a demo that showed how this is meant to be used, so I wrote a quick one, but maybe it's not how this is intended to be used. (I only found a demo that's intended to show that providing `useTres` doesn't break things)

alvarosabu:
> I grabbed the latest revision. provide from outside the <TresCanvas> doesn't seem to work on components within the <TresCanvas> – I tried adding an inject in SubComponentWithInject.vue.

> I could be wrong though. I didn't find a demo that showed how this is meant to be used, so I wrote a quick one, but maybe it's not how this is intended to be used. (I only found a demo that's intended to show that providing useTres doesn't break things)

Hi @andretchen0 yeah no worries I was just trying stuff, will set this PR to draft.

I did manage to make it work with `useTres` and such but I can't make it work with reactivity. The inject inside just get the value the first time and it doesn't get updated further (for example clicking the button outside)

I'm currently clueless how to make it work, I opened a help request here in vue discord https://discord.com/channels/325477692906536972/1283405486175031418

andretchen0:
> I'm currently clueless how to make it work

If we want to be explicit about `provide`s that are "bridged", a component like this works:

```vue
<script setup lang="ts">
import { provide } from 'vue'

interface Props {
  keysValues: Record<string, any>
}
const props = withDefaults(defineProps<Props>(), {
  keysValues: () => ({})
})
for (const [key, value] of Object.entries(props.keysValues)) {
  provide(key, value)
}
</script>

<template>
  <TresGroup>
    <slot></slot>
  </TresGroup>
</template>
```

```vue
 <TresCanvas>
    <TresPerspectiveCamera />
    <OrbitControls />

    <ProvideBridge :keys-values="{ foo: inject('foo') }">
        <Box /> <!-- Box can inject "foo" -->
     </ProvideBridge>
</TresCanvas>
```

src/components/TresCanvas.vue
```diff
@@ -104,14 +108,41 @@ const canvas = ref<HTMLCanvasElement>()
...
+      if (ctx) { ctx.app = instance?.appContext.app as App }
+      const provides = {}
+
+      // Helper function to recursively merge provides from parents
+      function mergeProvides(currentInstance: any) {
+        if (!currentInstance) { return }
+
+        // Extract provides from the current instance and merge them
+        if (currentInstance.provides) {
+          Object.assign(provides, currentInstance.provides)
```
andretchen0, line 127: Precedence is backwards here.

If the same key is provided more than once in the ancestor chain, the most distant ancestor is used. But it should be the closest.

Here's a set of files that shows the error.

* Save each code block to the filename preceding it.
* Open `index.vue`
* Open the console.

index.vue

```vue
<script setup lang="ts">
import { provide } from 'vue'
import TheExperience from './TheExperience.vue'

provide('test', '❌ Precendence is incorrect')
</script>

<template>
  <TheExperience />
</template>
```

TheExperience.vue

```vue
<script setup lang="ts">
import { provide } from 'vue'
import ComponentWithInject from './ComponentWithInject.vue'

provide('test', '✅ Precedence is correct')
</script>

<template>
  <TresCanvas>
    <ComponentWithInject />
  </TresCanvas>
</template>
```

ComponentWithInject.vue

```vue
<script setup lang="ts">
import { inject } from 'vue'

console.log(inject('test'))
</script>

<template>
  <TresMesh>
    <TresBoxGeometry />
    <TresMeshNormalMaterial />
  </TresMesh>
</template>
```

andretchen0, line 127: Precedence is backwards here.

If the same key is provided more than once in the ancestor chain, the most distant ancestor is used. But it should be the closest.

Here's a set of files that shows the error.

* Save each code block to the filename preceding it.
* Open `index.vue`
* Open the console.

index.vue

```vue
<script setup lang="ts">
import { provide } from 'vue'
import TheExperience from './TheExperience.vue'

provide('test', '❌ Precendence is incorrect')
</script>

<template>
  <TheExperience />
</template>
```

TheExperience.vue

```vue
<script setup lang="ts">
import { provide } from 'vue'
import ComponentWithInject from './ComponentWithInject.vue'

provide('test', '✅ Precedence is correct')
</script>

<template>
  <TresCanvas>
    <ComponentWithInject />
  </TresCanvas>
</template>
```

ComponentWithInject.vue

```vue
<script setup lang="ts">
import { inject } from 'vue'

console.log(inject('test'))
</script>

<template>
  <TresMesh>
    <TresBoxGeometry />
    <TresMeshNormalMaterial />
  </TresMesh>
</template>
```

andretchen0 (changes requested):

src/components/TresCanvas.vue
alvarosabu: Sorry, you are right @andretchen0  I thought I fixed it. [aa1dce5](https://github.com/Tresjs/tres/pull/806/commits/aa1dce5791f7556eba977c7e774e3bb118501c5b) should do the trick

docs/api/tres-canvas.md
```diff
@@ -92,7 +92,7 @@ renderer.shadowMap.type = PCFSoftShadowMap
 | **toneMappingExposure** | Exposure level of tone mapping. | `1` |
 | **useLegacyLights** | Whether to use the legacy lighting mode or not | `true` |
 | **windowSize** | Whether to use the window size as the canvas size or the parent element. | `false` |
-| **disableProvideBridge** | Disable the provide/inject bridge between Vue context and TresCanvas. | `false` |
+| **enableProvideBridge** | Disable the provide/inject bridge between Vue context and TresCanvas. | `false` |
```
andretchen0, line 95: The description and the default value didn't get updated with the property name. It should probably be something like:

```md
| **enableProvideBridge** | Enable the provide/inject bridge between Vue context and TresCanvas. | `true` |
```

andretchen0 (changes requested):

andretchen0 (approved):

PR #820, opened 2024-09-05
ci: add release action

alvarosabu:


.github/workflows/release.yml
```diff
@@ -0,0 +1,37 @@
...
+        with:
+          fetch-depth: 0
+      - name: Setup pnpm
+        uses: pnpm/action-setup@v4
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v4
+      - name: Initialize Git User
+        run: |
+          git config user.name "${GITHUB_ACTOR}"
+          git config user.email "alvaro.saburido@gmail.com"
```
colinscz, line 26: My suggestion would be to pack the user.email also in an env variable or secret, that way if in the future you want to change it or maybe us an organisation email (github(at).tresjs.org) or something else you can easily change it without having to update the code.

.github/workflows/release.yml
```diff
@@ -0,0 +1,37 @@
+name: Release & Publish to NPM
+
+on:
+  pull_request:
+  push:
+    branches: [main]
```
colinscz, line 6: I am not sure I guess you don't want to always execute a full release when a merge on master is done, probably something like an insider release with a hash could be an idea. 

We have implemented a workflow in our project that creates an insider-release after every merge to main. It's far from perfect but could be an option: 
https://github.com/six-group/six-webcomponents/blob/main/.github/workflows/insider-release.yml

Issue #801, opened 2024-08-17
`pointer-{leave,out}` fired multiple times on single "leave"


andretchen0:
### Describe the bug

An object's `pointer-leave` event is fired more than once on a single "leave" if it is the descendant of another element that also defines a `pointer-leave`:

```vue
<TresGroup @pointer-leave="() => {}">
  <TresMesh @pointer-leave="() => { console.log('Leave') }">
    <TresBoxGeometry :args="[1, 1, 1]" />
    <TresMeshToonMaterial color="#efefef" />
  </TresMesh>
</TresGroup>
```

See also: `pointer-out` exhibits similar behavior.

### Reproduction

https://stackblitz.com/edit/tresjs-minimal-reproduction-qd169j?file=src%2Fcomponents%2FTheExperience.vue

### Steps to reproduce

* Open Stackblitz link in a browser.
* Mouse over cube, then leave.
* See browser console: multiple `pointer-leave` events will be logged.

### Used Package Manager

npm

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/Tresjs/tres/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/Tresjs/tres/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://tresjs.org/guide).
- [X] Check that there isn't [already an issue](https://github.com/tresjs/tres/issues) that reports the same bug to avoid creating a duplicate.
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

Tinoooo:
This is fixed with v5.

PR #831, opened 2024-09-12
refactor: events (WIP)

andretchen0:
* (fix) BREAKING - remove `event.intersects` – use `event.intersections`
* (fix) BREAKING - remove `event.stopPropagating` – use `event.stopPropagation`
* (fix) BREAKING – from TresCanvas, don't emit pointer/click events bubbled from Tres objects
* (fix) BREAKING – `event.delta` is now reset to `0` following a `click`.
* (fix) BREAKING – `pointerleave` handlers were sent previous intersections at `event.intersections`, now current intersections.
* (fix) BREAKING – `@dblclick` now fires whenever the canvas `dblclick` is emitted. Any TresObject with an `@dblclick` handler will receive the event if it is under the pointer. (Previously objects not under the initial click did not receive the `@dblclick`.)
* (fix) `event.pointer` is now defined – previously typed but was `undefined` in implementation.
* (fix) `event.unprojectedPoint` is now properly calculated. Was previously left as `TODO`.
* (fix) `event.eventObject` holds the object that registered the event handler. Was previously typed but `undefined` in implementation.
* (fix) use Vue-style ("flatcase") event names, e.g. `@pointerdown`. Keep support for old-style ("kebab-case") event names, e.g. `@pointer-down`, and warn. Mixing both on the same object is not supported and may lead to handlers being overwritten.
* (fix) `@pointer{leave,out}` was fired multiple times on a single "leave" – Issue #801
* (feat) BREAKING – `pointermissed` fires when the object that registered the handler is missed – previously only fired when ALL objects were missed
* (feat) add filtering/sorting for intersections, prior to handling events. `:events={filter: (intersections) => ...}`
* (feat) support Vue event modifiers – `stop, prevent, self, once`. (Tres cannot support `passive` and does not currently support `capture`.)
* (feat) via `:events-target` prop, allow users to specify an HTML `addEventListener` target – allows Tres to respond to events, even if there's an overlay on the canvas.
* (feat) via `:events-enabled` prop, allow `eventManager` to be enabled/disabled while Tres is running.
* (feat) via `:events` prop, allow `eventManager` functions to be set/overwritten (non-reactive).
* (feat) via `:events` prop, allow events system to be disabled completely: `<TresCanvas :events="null" />` (non-reactive)
* (feat) `pointer{leave,out}` is triggered (if necessary) just prior to object removal.
* (feat) `:blocking="true"` makes a subtree "solid"; objects behind objects in the subtree will not be "hit".
* (perf) meshes without events on self/ancestors are no longer hit-tested by default.
* (refactor) `context.eventManager` now creates its own raycaster(s) or no raycaster – depending on needs. It does not use `context.raycaster`.
* (refactor) BREAKING – remove the `useRaycaster` composable.
* (refactor) BREAKING – remove the `useEventManager` composable.
* (refactor) `eventsManager` is dependency-inverted. Users can set/modify its internal functions.


dghez:
@andretchen0 // @alvarosabu can't see anything in the docs/master referencing to this refactor. 
I need a way to disable every event coming from the raycaster because if i have a `TresPerspectiveCamera` and then i unmount it, the raycaster starts to throw some errors like the camera doesn't exists, even if never use the raycaster system.



image
image

PR #837, opened 2024-09-17
chore: coverage setup

alvarosabu:


vite.config.ts
```diff
@@ -44,9 +44,13 @@ export default defineConfig({
     Inspect(),
   ],
   test: {
```
andretchen0, line 46: With the setup in this PR, 0 tests are run after `pnpm test:ui`.

Tests run using the following config, which also excludes `playground`, `docs` and `**/sponsorkit**/**` from coverage:

```
import { coverageConfigDefaults } from 'vitest/config'
```

```
  test: {
    environment: 'jsdom',
    globals: true,
    coverage: {
      provider: 'v8',
      exclude: [
        ...coverageConfigDefaults.exclude,
        'playground/**',
        'docs/**',
        '**/sponsorkit**/**',
      ],
    },
  },
```

Taken from here: [Vitest coverage.exclude docs](https://vitest.dev/config/#coverage-exclude)

andretchen0 (changes requested):
Hey @alvarosabu ,

Thanks for looking into this. 

For me, after adding the changes here, no tests run. 

I've appended changes to `vite.config.ts` that work for me locally.

Do they work for you?

vite.config.ts
alvarosabu: Hi @andretchen0 for me they work, but I updated the config with your suggestion

alvarosabu:
Hi @andretchen0 I added your suggestion, is it good to merge?

PR #838, opened 2024-09-22
chore: remove obsolete an non-functional prop disableRender

Tinoooo:


alvarosabu:
Thanks for giving me a hand with this @Tinoooo

PR #850, opened 2024-10-04
support Symbol keys in provide/inject (fix #849)

### LINKED ISSUE #849: Provide/Inject from Vue component not working with Symbol keys
### Describe the bug

First of all, thanks @alvarosabu for implementing #806.

I went to try it out today, but it wasn't working. After some debugging, I realized it's because we are using [Symbol keys](https://vuejs.org/guide/components/provide-inject#working-with-symbol-keys) for many of our provides.

`Object.entries` only includes string-keyed properties, so the Symbol-keyed properties are not included.

Either `Object.getOwnPropertySymbols` or `Reflect.ownKeys` should be able to at least get the symbol keys.

### Reproduction

https://stackblitz.com/edit/tresjs-basic-wkbbis?file=src%2FApp.vue,src%2Fcomponents%2FTheExperience.vue,src%2Fcomponents%2FSubComponent.vue,package.json

### Steps to reproduce

Run the Stackblitz project (`npm install && npm run dev`) if not automatically run.

The console will have the following log messages:

```
TheExperience.vue string-keyed string-keyed
TheExperience.vue symbol-keyed symbol-keyed
```

These are the values of two injections provided in `App.vue`.

Clicking the box logs the following:

```
SubComponent.vue string-keyed string-keyed
SubComponent.vue symbol-keyed undefined
```

As the symbol-keyed provision can not be read inside the `TresCanvas`.

### System Info

_No response_

### Used Package Manager

npm

### Code of Conduct

eriklumme1:
As `Object.entries(...)` only lists string properties, and Vue supports provide/inject [with Symbols as keys](https://vuejs.org/guide/components/provide-inject#working-with-symbol-keys), all provisions are not picked up.

I opted for `Reflect.ownKeys`, as its a good choice per the MDN documentation:

[Object.entries()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries)

 > The Object.entries() static method returns an array of a given object's **own enumerable string-keyed property** key-value pairs.

[Reflect.ownKeys()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Reflect/ownKeys)

> The Reflect.ownKeys() static method returns an array of the target object's **own property keys**

> Return Value
> Array of the target object's own property keys, **including strings and symbols**. For most objects, the array will be in the order of:

I've modified the #732 example to add the symbols. Prior to my change, this is the output:

Screenshot 2024-10-04 at 15 10 29

And with my changes it's this

Screenshot 2024-10-04 at 15 19 37

### DIFF
--- a/playground/vue/src/pages/issues/732/SubComponentWithInject.vue
+++ b/playground/vue/src/pages/issues/732/SubComponentWithInject.vue
@@ -1,15 +1,19 @@
 <!-- eslint-disable no-console -->
 <script setup lang="ts">
 import { inject, watchEffect } from 'vue'
+import { injectSymbol } from './constants'
 
 const awiwi = inject('awiwi')
 const bululu = inject('bululu')
+const symbolInjection = inject(injectSymbol)
 const vRoute = inject('v-route')
 console.log(inject('test'))
 
 watchEffect(() => console.log('tres:awiwi', awiwi?.a))
 
 watchEffect(() => console.log('tres:bululu', bululu?.value))
+
+watchEffect(() => console.log('tres:symbolInjection', symbolInjection?.value))
 
 watchEffect(() => console.log('tres:v-route', vRoute))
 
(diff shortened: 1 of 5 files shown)


alvarosabu:
@eriklumme1 thanks a lot for raising this and providing a working solution. You are awesome

eriklumme1:
@alvarosabu thanks a lot for getting this in!

PR #870, opened 2024-11-11
docs(attach): rewording

andretchen0:
This PR makes some minor rewording fixes to the `attach` docs and playgrounds.


alvarosabu (approved):
Thanks for correcting this, is more clear now

PR #885, opened 2024-12-19
fix: node16 types

userquin:
Check changes here https://github.com/Tresjs/tres/pull/885/files#diff-5fd41963f486d1796606e250bceaafc58a425d5b924c22f98dafaae17de208d3R26 , just removed some error there (the option can be undefined)

With this PR applying same strategy used here https://github.com/Tresjs/post-processing/pull/152:

image


alvarosabu:
@userquin will close this in favor of #888

PR #916, opened 2025-02-05
fix: remove camera warning log on default camera creation

alvarosabu:


andretchen0 (approved):
Great! Glad to have less noise in the console.

PR #918, opened 2025-02-05
fix: improve Object3D disposal method

### LINKED ISSUE #912: Helpers disposal issue
### Describe the bug

Whenever we unmount the TresCanvas on route change, for example, we are recursively disposing and deleting scene references. This seems to interfere with Helper's own disposal cycle.

```
chunk-TMRGR37B.js?v=d91265db:20290 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'dispose')
    at DirectionalLightHelper.dispose (chunk-TMRGR37B.js?v=d91265db:20290:30)
    at disposeObject3D (index.ts:302:14)
    at index.ts:294:29
    at Array.forEach (<anonymous>)
    at disposeObject3D (index.ts:294:12)
    at dispose (TresCanvas.vue:162:3)
    at unmountCanvas (TresCanvas.vue:183:3)
```

### Reproduction

Internal

### Steps to reproduce

1. Add a helper to the scene
2. Change route.

### System Info

```shell
System:
    OS: macOS 15.2
    CPU: (8) arm64 Apple M1 Pro
    Memory: 192.69 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.11.0 - /usr/local/bin/node
    npm: 10.2.4 - /usr/local/bin/npm
    pnpm: 9.15.4 - ~/Library/pnpm/pnpm
    bun: 1.0.2 - ~/.bun/bin/bun
  Browsers:
    Brave Browser: 120.1.61.116
    Chrome: 132.0.6834.160
    Safari: 18.2
  npmPackages:
    @tresjs/cientos: 4.0.3 => 4.0.3 
    @tresjs/core: workspace:^ => 4.3.2 
    @tresjs/leches: https://pkg.pr.new/@tresjs/leches@9ad0cd3 => 0.14.1
```

### Used Package Manager

pnpm

### Code of Conduct

alvarosabu:
- Remove unnecessary delete operations on mesh geometry and material
- Add TODO comment for Scene-specific cleanup
- Simplify material and geometry disposal logic

Closes #912


andretchen0 (commented):
This looks good to me. Thanks!

When the memory tests are merged to `main`. I'll have a closer look.

src/utils/index.ts
```diff
@@ -294,7 +294,7 @@ export function disposeObject3D(object: TresObject): void {
   children.forEach(child => disposeObject3D(child))
 
   if (object instanceof Scene) {
-    // Optionally handle Scene-specific cleanup
+    // TODO: Handle Scene-specific cleanup
```
andretchen0, line 297: This doesn't appear to be the place to handle scene memory leaks, as far as I can tell. 

As of this PR, the scenes are retained via `useTres` and `nodeOps.createComment`.

andretchen0 (approved):
Looks good. Just one non-blocking comment.

src/utils/index.ts
alvarosabu: @andretchen0 are we entirely sure?  setting the scene to null here did improve the test we did in the closed PR, most likely we will need to tackle the memory leaks here and  via useTres and nodeOps.createComment.

andretchen0: It's not that setting `scene = null` or something similar won't impact the memory numbers. 

The problem with that approach is that:

* it addresses the *effect* of the leak and not its *cause*. 
* it addresses the *effect* of the leak in a way that makes the program buggier and less predictable.

## Cause of the leak

In the abstract, we have something like:

```js
const myRetainer = { a: { b: { c: { scene: { children: [...], ... } } } } }
```

If we do ...

```js
myRetainer.a.b.c.scene = null
```

... then we're fixing the **effect** of the leak and memory numbers will improve. But this isn't the **cause** of the leak.

The *cause* of the leak is that we're holding on to `myRetainer`, e.g., in `useTres`, so it can't be garbage collected. If we find and remove the references to `myRetainer`, `scene` will be garbage collected along with everything else that's only retained by `myRetainer`.

And – importantly – by addressing the *cause* of the leak, we'll also likely avoid bugs like this PR is fixing. 

## Bugs caused by addressing leak *effects*

In trying to address the *effect* of a memory leak, this was added to the codebase:

```js
delete mesh.geometry
```

But THREE helper disposal expects `mesh.geometry` to be defined, and when it wasn't, it threw.

In a similar vein, we address the *effect* of a memory leak by setting `scene = null`, But then any code executed after that point that calls `scene.[some method]()` will throw. 

If we address the *cause* of the leak, `scene` will be garbage collected without setting it to `null`, so there's no risk of producing that sort of bug.

PR #928, opened 2025-02-10
Remove unescessary call to getElapsedTime

AlaricBaraou:
The call is not causing any bug since it is called immediatly after `getDelta` but it is doing a duplicate job with `getDelta()` that is already updating the `.elapsedTime` property.

https://github.com/mrdoob/three.js/blob/master/src/core/Clock.js#L58

### DIFF
--- a/src/composables/useRenderLoop/index.ts
+++ b/src/composables/useRenderLoop/index.ts
@@ -37,7 +37,7 @@ const { pause, resume, isActive } = useRafFn(
 
 onAfterLoop.on(() => {
   delta = clock.getDelta()
-  elapsed = clock.getElapsedTime()
+  elapsed = clock.elapsedTime
 })
 
 let startedOnce = false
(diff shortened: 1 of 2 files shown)


colinscz:
@AlaricBaraou could you by chance rebase your branch on the latest status of main? :) 

Thanks!

AlaricBaraou:
Thanks @colinscz, it should be good now 👍

colinscz:
Thanks @AlaricBaraou 
I just link the documentation entry here as well. https://threejs.org/docs/#api/en/core/Clock.getDelta 

@alvarosabu the PR is all yours :)

Tinoooo:
@AlaricBaraou Thank you for pointing to this. 🙌 I applied your suggestion to the changes we made concerning v5 [here](https://github.com/Tresjs/tres/pull/1046).

Tinoooo:
I'm closing this PR as the changes are reflected with the release of v5 now.

PR #954, opened 2025-03-07
Fix import of vLightHelper

tobiaswaelde:
Changed the import of `vLightHelper` from `@tresjs/cientos` to `@tresjs/core`. (as seen in the untranslated documentation)


alvarosabu:
Thanks @tobiaswaelde

PR #958, opened 2025-03-17
docs: add contribution guide link to docs and for different languages

### LINKED ISSUE #955: [Docs]: add Contribution guidelines to docs
### Description

Add https://github.com/Tresjs/.github/blob/main/CONTRIBUTING.md to the docs

### Suggested solution

Create a new page in your docs directory (e.g., contributing.md)
Use the VitePress remote markdown inclusion feature to fetch and embed the content directly from GitHub:

```md
# Contributing to TresJS

<script setup>
import { onMounted, ref } from 'vue'

const content = ref('')

onMounted(async () => {
  const response = await fetch('https://raw.githubusercontent.com/Tresjs/.github/main/CONTRIBUTING.md')
  content.value = await response.text()
})
</script>

<div v-html="content" class="remote-markdown"></div>
```

Add this page to your navigation sidebar in the VitePress config

If you need more customization or if the above approach doesn't work well with your setup, you could:

Create a build script that pulls the content from GitHub and saves it as a local file during build time
Use a GitHub workflow to automatically sync the file when changes are made
Simply copy the content manually and set up a reminder to update it periodically

Would you like more specific implementation details for any of these approaches?

### Alternative

_No response_

### Additional context

_No response_

### Validations

colinscz:
closes #955

I took inspiration from the Vitejs documentation where they added a link in the resources menu to their contribution markdown file and did the same here.

ps: I did the translation to chinese with chatgpt so I am not sure if it's grammatically correct.

~~PR for removing contributing markdown from: [.github tresjs](https://github.com/Tresjs/.github/pull/4)~~


CONTRIBUTING.md
```diff
@@ -1,19 +1,83 @@
-![repository-banner.png](https://res.cloudinary.com/alvarosaburido/image/upload/v1683452574/repo-banner_d2xeem.png)
+# Contributing Guide
```
colinscz, line 1: @alvarosabu I did a simple replace for now, but happy to adjust the document here and introduce repository banner and other paragraphs again which were previously there. :)

colinscz: I will do a separate commit with your suggested solution in the raised issue, so we have two options to consider :)

docs/.vitepress/config/en.ts
```diff
@@ -127,6 +127,10 @@ export const enConfig: LocaleSpecificConfig<DefaultTheme.Config> = {
             text: 'Issues',
             link: 'https://github.com/Tresjs/tres/issues',
           },
+          {
+            text: 'Contributing',
+            link: 'https://github.com/Tresjs/tres/blob/main/CONTRIBUTING.md',
```
alvarosabu, line 132: @colinscz should be https://github.com/Tresjs/.github/blob/main/CONTRIBUTING.md

alvarosabu (changes requested):

CONTRIBUTING.md
colinscz: I see, so if I understand you correctly you would prefer to keep the contribution markdown in the .github repository if we go for this approach and link their and remove the duplicate one in this repository. 
Also fine for me :) 
I will adjust the code before I do have a look at the other approach.

docs/.vitepress/config/en.ts
colinscz: addressed in de8a2a660a37e560c202e8af25258a6f95e2ec24

CONTRIBUTING.md
colinscz: addressed in de8a2a660a37e560c202e8af25258a6f95e2ec24

docs/contribute/contribution.data.js
```diff
@@ -0,0 +1,13 @@
+import { createMarkdownRenderer } from 'vitepress'
+
+export default {
+
```
colinscz, line 4: I pushed a solution based on the following links and suggest approaches I found for the english version so you have an example:

- https://vitepress.dev/guide/data-loading
- https://github.com/vuejs/vitepress/discussions/3049
- https://github.com/vuejs/vitepress/discussions/2921#discussioncomment-7023589 

But I am not really convinced and as of now it seems to me a bit of an overkill to introduce a data loader for this one markdown file and an extra section. 

Wdyt @alvarosabu ? 

I suggest we stick with the other approach that Vitepress is already using. :)

docs/contribute/contributing.md
```diff
@@ -0,0 +1,16 @@
+# Contributing to TresJS
```
alvarosabu, line 1: We should probably remove this h1 here since the contribution markdown already has one @colinscz 
Screenshot 2025-05-13 at 15 19 21

CONTRIBUTING.md
alvarosabu, line 1: Maybe we should remove this one in favor of https://github.com/Tresjs/.github/blob/main/CONTRIBUTING.md? @colinscz

docs/contribute/contributing.md
colinscz: Sure @alvarosabu - I just removed it. Can I assume you favor the approach in this commit: e66709359e659d12b7a2a95946795bcd76ef62ee with the data renderer? (See other open discussion)

CONTRIBUTING.md
alvarosabu: I think we don't have specific ones, they are standard to the ecosystem, lets remove it

colinscz: a4225016f22b329f23017a4ec206bc63fac12296

alvarosabu (approved):

PR #970, opened 2025-03-28
feat: 962-createrenderer-composable

alvarosabu:
BREAKING CHANGE: This commit introduces significant changes to the renderer system:

- Removes renderer presets in favor of direct configuration
- Makes useTresContextProvider async
- `useRenderer` now only returns the renderer from context.
- Changes default renderer values
- Adds support for multiple renderer types (WebGPU, CSS2D, CSS3D, SVG) via prop callback
- Improves TypeScript types and documentation
- Refactors renderer creation and setup into separate modules (internal)


src/composables/useTresContextProvider/index.ts
```diff
@@ -83,20 +85,20 @@ export interface TresContext {
...
+export async function useTresContextProvider({
   scene,
   canvas,
   windowSize,
   rendererOptions,
   emit,
 }: {
   scene: TresScene
-  canvas: MaybeRef<HTMLCanvasElement>
+  canvas: Ref<HTMLCanvasElement>
```
Tinoooo, line 96: **question:** Why can't this stay a `MaybeRef`?

src/composables/useRenderer/index.ts
```diff
@@ -1,260 +1,5 @@
...
-
-import { normalizeColor } from '../../utils/normalize'
-import { logError } from '../../utils/logger'
-import { rendererPresets } from './const'
-
-type TransformToMaybeRefOrGetter<T> = {
-  [K in keyof T]: MaybeRefOrGetter<T[K]> | MaybeRefOrGetter<T[K]>;
-}
-
-export interface UseRendererOptions extends TransformToMaybeRefOrGetter<WebGLRendererParameters> {
```
Tinoooo, line 28: **suggestion:** I liked the fact, that the interface was defined here since this is the entity that needs these values. If I get this right now, `TresCanvasProps` is defined because it uses `createRenderer`. I think defining the types close to where the dependencies arise is a better practice for readability and mantainability reasons. Therefore I suggest moving the interface definition.

src/core/createRenderer.ts
```diff
@@ -0,0 +1,23 @@
+import { WebGLRenderer } from 'three'
+import * as is from '../utils/is'
+import type { TresContext } from '../composables'
+import type { TresRenderer } from '../types'
+import type { TresCanvasProps } from '../components/TresCanvas.vue'
+
+export async function createRenderer(ctx: TresContext, options: TresCanvasProps): Promise<TresRenderer> {
```
Tinoooo, line 7: **suggestion:** I suggest explicitly naming the context dependencies as we did before (`contextParts`). Doing so shows the dependencies of each composable or creator and allows for a nice orchestration in `useTresContextProvider`.

src/core/createRenderer.ts
```diff
@@ -0,0 +1,23 @@
...
+  if (options.renderer) {
+    const fnOrRenderer = options.renderer
+    if (typeof fnOrRenderer === 'function') {
+      return await fnOrRenderer(ctx)
+    }
+    return fnOrRenderer
```
Tinoooo, line 13: **issue:** There is an error in the typing. `fnOrRenderer` is of type `never` at this point.

src/core/setupRenderer.ts
```diff
@@ -0,0 +1,163 @@
...
+  clearAlpha: {
+    set: (renderer, value) => renderer.setClearAlpha(value as number),
+    immediate: true,
+  },
+}
+
+// Modified setup function to handle both types of properties
+export function setupWebGLRenderer(
+  initialRenderer: WebGLRenderer,
+  options: Record<string, any>,
```
Tinoooo, line 74: **suggestion:** The options should be very clear. This seems very loosely typed. I suggest typing this strictly.

src/core/setupRenderer.ts
```diff
@@ -0,0 +1,163 @@
...
+    },
+    immediate: true,
+  },
+  clearAlpha: {
+    set: (renderer, value) => renderer.setClearAlpha(value as number),
+    immediate: true,
+  },
+}
+
+// Modified setup function to handle both types of properties
```
Tinoooo, line 71: **question:** What `properties` are meant here? Should this be `renderers`?

src/composables/useTresContextProvider/index.ts
```diff
@@ -130,24 +133,14 @@ export function useTresContextProvider({
...
-    },
-  )
-
   const ctx: TresContext = {
     sizes,
     scene: localScene,
     camera,
     cameras: readonly(cameras),
-    renderer,
+    renderer: shallowRef(null as unknown as Renderer),
```
Tinoooo, line 141: **suggestion:** This is `null` is not a `Renderer`. I suggest not using this type assertion. It seems hacky to me,

src/core/setupRenderer.ts
```diff
@@ -0,0 +1,163 @@
...
+  }, {
+    immediate: true,
+  })
+
+  // Watch properties that need setter methods
+  Object.entries(rendererPropertyHandlers).forEach(([key, handler]) => {
+    watch(
+      () => options[key],
+      (value) => {
+        if (value === undefined) { return }
```
Tinoooo, line 106: **question:** I haven't tested, but the code gives me the impression, that unsetting values to the renderers default, when props change to undefined, does not work anymore as it worked before. Is this intentionally dropped?

src/composables/useTresContextProvider/index.ts
```diff
@@ -209,6 +210,8 @@ export function useTresContextProvider({
   })
 
   onUnmounted(() => {
+    ctx.renderer.value.dispose()
+    ctx.renderer.value.forceContextLoss()
```
Tinoooo, line 214: **suggestion:** This should be in the responsibility of the renderer's composable.

src/core/setupRenderer.ts
```diff
@@ -0,0 +1,163 @@
...
+  immediate?: boolean
+}
+
+interface DirectProperty {
+  key: keyof WebGLRenderer | 'shadowMap.enabled' | 'shadowMap.type' | 'physicallyCorrectLights'
+  immediate?: boolean
+}
+
+// Properties that can be set directly on the renderer
+const directProperties: Record<string, DirectProperty> = {
```
Tinoooo, line 20: **question:** Could this be typed so that there is a function that returns the element that should be watched? I think this would allow for much better type safety and the dot notation handling might not be required.

Tinoooo (changes requested):

PR #976, opened 2025-04-03
fix: linter findings and force linter to be executed on every push

colinscz:
I noticed that the linter check was executed but only on master and some of my linter findings were not checked. 🤔 

I fixed them in this PR and changed the workflow to be executed on every push to a PR or to main. 
The same syntax is also on the test workflow, maybe that one needs also that change. Happy to address this as well. 
I circled back to the last PR and couldn't find an execution also not in the actions tab for that PR, which was very confusing considering the trigger was for all kinds of "push". 👀 

Maybe the Github action syntax was outdated, not sure though.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,5 +1,11 @@
 name: Run linters
-on: [push]
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
 
 env:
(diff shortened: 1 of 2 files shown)


Tinoooo:
I'm closing this as since the linter will be called together with the tests and typecheck in [the next major version](https://github.com/Tresjs/tres/blob/next/.github/workflows/ci.yml). Thank you @colinscz tough!

PR #988, opened 2025-04-16
chore: type checking in ci

Tinoooo:


.github/workflows/typecheck.yml
```diff
@@ -0,0 +1,28 @@
+name: Run typecheck
```
Tinoooo, line 1: **suggestion to myself:** Let's combine typecheck and lint workflow to reduce github action minutes.

Tinoooo, line 1: **suggestion to myself:** Let's combine typecheck and lint workflow to reduce github action minutes.

alvarosabu (approved):

PR #991, opened 2025-04-17
refactor: reorganize devtools setup and imports

alvarosabu:
- Moved the `registerTresDevtools` import in `TresCanvas.vue` to improve clarity and ensure proper registration of devtools.
- Renamed `setupDevtools` to `setupTresDevtools` in `setupDevtools.ts` for consistency with the new naming convention.
- Updated the `registerTresDevtools` function in `plugin.ts` to call the renamed `setupTresDevtools`, ensuring proper integration of the devtools setup.
- Removed the unused `setupDevtools` call from `useTresContextProvider` to streamline the context provider logic.

### DIFF
--- a/src/devtools/setupDevtools.ts
+++ b/src/devtools/setupDevtools.ts
@@ -4,7 +4,7 @@ import type { TresContext } from '../composables'
 import type { TresObject } from '../types'
 import { onUnmounted } from 'vue'
 
-export function setupDevtools(ctx: TresContext) {
+export function setupTresDevtools(ctx: TresContext) {
   if (!ctx) { return }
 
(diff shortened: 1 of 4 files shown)


src/devtools/setupDevtools.ts
Tinoooo, line 1: **suggestion(non-blocking):** I suggest moving this to the composables folder.

Tinoooo (approved):

src/devtools/setupDevtools.ts
alvarosabu: I'm thinking on a better name already. when I move it I will call it from TresCanvas instead then

PR #1000, opened 2025-04-23
refactor!: removed useTresReady, added isReady to the renderer in the…

Tinoooo:
… context

BREAKING CHANGE:

- useTresReady is no longer available, it has been replaced by isReady in the renderer in the context
- onTresReady is no longer available. renderer.isReady should be leveraged instead


alvarosabu (approved):
LGTM @Tinoooo thanks for taking care of it

PR #1004, opened 2025-04-27
refactor!: context camera is now a state

### LINKED ISSUE #985: Add `useCamera` return object to state instead of spreading it
### Description

As a maintainer of TresJS, I would like to have a more organized and less polluted TresContext.

```ts
export interface TresContext {
  ...
  camera: ComputedRef<Camera | undefined>
  cameras: DeepReadonly<Ref<Camera[]>>
  registerCamera: (maybeCamera: unknown) => void
  setCameraActive: (cameraOrUuid: Camera | string) => void
  deregisterCamera: (maybeCamera: unknown) => void
  eventManager?: TresEventManager
 ...
}
```

### Suggested solution

```ts
export interface TresContext {
  ...
  camera: useCamera()
 ...
}


### Alternative

_No response_

### Additional context

_No response_

### Validations

### PR DESCRIPTION
**Pull request by alvarosabu**
BREAKING CHANGE: camera ctx property is now an object with the camera manager instead of the active camera

Closes #985

### CODE REVIEW

File: playground/vue/src/pages/cameras/index.vue [hunk 1]
```diff
@@ -1,82 +1,58 @@
...
 
-watch(() => [cameraType.value.value, manual.value.value], () => {
-  state.cameraType = cameraType.value.value
-  if (cameraType.value.value === 'perspective') {
-    state.camera = new PerspectiveCamera(75, 1, 0.1, 1000)
-  }
-  else if (cameraType.value.value === 'orthographic') {
-    state.camera = new OrthographicCamera(-1, 1, 1, -1, 0, 1000)
-    state.camera.zoom = 20
+watch(cameraType, (newCameraType) => {
```
Tinoooo, line 40: **suggestion(non-blocking):** This should be replaced by a computed.

src/core/nodeOps.ts
```diff
@@ -114,7 +115,10 @@ export const nodeOps: (context: TresContext) => RendererOptions<TresObject, Tres
       context.eventManager?.registerObject(child)
     }
 
-    context.registerCamera(child)
+    if (isCamera(child)) {
+      context.camera?.registerCamera(child as TresCamera)
```
Tinoooo, line 119: **suggestion:** This type assertion should not be required as it comes right after a type guard.

src/core/nodeOps.ts
```diff
@@ -339,6 +343,10 @@ export const nodeOps: (context: TresContext) => RendererOptions<TresObject, Tres
     else if (!target.isColor && target.setScalar) { target.setScalar(value) }
     else { target.set(value) }
 
+    if (node.isCamera) {
+      node.updateProjectionMatrix()
```
Tinoooo, line 347: **question:** Why is this required?
**suggestion:** A type guard should be used so we can be sure `updateProjectionMatrix` exists on `node`.

src/core/nodeOps.ts
```diff
@@ -76,6 +76,7 @@ export const nodeOps: (context: TresContext) => RendererOptions<TresObject, Tres
 
     if (!obj) { return null }
 
+    // Opinionated default to avoid user issue not seeing anything if camera is on origin
     if (obj.isCamera) {
```
Tinoooo, line 80: **suggestion:** Type guard should be used.

Tinoooo (changes requested):

src/composables/useCamera/index.ts
```diff
@@ -1,85 +1,151 @@
...
+        // Update orthographic camera
+        else if (isOrthographicCamera(camera)) {
+          // Use a fixed frustum size for better visualization
+          const frustumSize = 10
+          const aspect = sizes.aspectRatio.value
+
+          camera.left = frustumSize * aspect / -2
+          camera.right = frustumSize * aspect / 2
+          camera.top = frustumSize / 2
+          camera.bottom = frustumSize / -2
```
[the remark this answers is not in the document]
alvarosabu: @Tinoooo did you check the playground what happens if you set these changes?

Screenshot 2025-04-29 at 20 22 58

From the threejs journey, you need to set a fixed frustum and use the aspect ratio only in the left and right (first and second parameter) This keeps the same “zoom” level, but adjusts the width to match the aspect ratio.

Screenshot 2025-04-29 at 20 29 01

src/core/nodeOps.ts
alvarosabu: @Tinoooo when changing props, like near, far, or anything that affects the projection matrix you need to tell the camera to update it manually.

src/composables/useCamera/index.ts
```diff
@@ -1,85 +1,151 @@
...
-  // the computed does not trigger, when for example the camera position changes
-  const cameras = ref<Camera[]>([])
-  const camera = computed<Camera | undefined>(
-    () => cameras.value[0],
+/**
+ * Interface for the return value of the useCamera composable
+ */
+export interface UseCameraReturn {
+  /**
+   * The active camera
```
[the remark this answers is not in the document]
alvarosabu: mmmm @Tinoooo I get the part about the duplication but if I try what you suggest I get this

Screenshot 2025-04-30 at 17 57 12

So I infer that it was because the composable uses the type to define the return here:

```ts
export const useCameraManager = ({ sizes }: UseCameraParams): UseCameraReturn => {
```

If I remove it then it no longer shows the text of the annotation 

Screenshot 2025-04-30 at 17 59 45

src/composables/useCamera/index.ts
```diff
@@ -1,85 +1,151 @@
...
+
+          camera.left = frustumSize * aspect / -2
+          camera.right = frustumSize * aspect / 2
+          camera.top = frustumSize / 2
+          camera.bottom = frustumSize / -2
+
+          // Ensure the camera is at a good position to see the scene
+          if (!camera.position.z) {
+            camera.position.z = 10
           }
```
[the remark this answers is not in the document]
Tinoooo: This seems to be fixed now.

src/composables/useCamera/index.ts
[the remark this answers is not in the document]
Tinoooo: We agreed to not adjust to the orthographic camera on size changes at all.

src/core/nodeOps.ts
```diff
@@ -114,7 +115,10 @@ export const nodeOps: (context: TresContext) => RendererOptions<TresObject, Tres
       context.eventManager?.registerObject(child)
     }
 
-    context.registerCamera(child)
+    if (isCamera(child)) {
+      context.camera?.registerCamera(child)
+      context.camera?.setActiveCamera(child.uuid)
```
Tinoooo, line 120: **suggestion:** This would mean that each added camera would automatically set active. As a user, I would not expect this. It can be especially cumbersome when using `v-if`. I suggest omitting this line.

src/core/nodeOps.ts
```diff
@@ -288,10 +292,31 @@ export const nodeOps: (context: TresContext) => RendererOptions<TresObject, Tres
...
+              (prevNode as unknown as Record<string, unknown>)[key] = newInstance[key]
+            }
+            catch (e) {
+              // Skip if property can't be set
+              console.warn(`Could not set property ${key} on ${instanceName}:`, e)
+            }
+          }
+        })
+
+        root = prevNode
```
Tinoooo, line 319: **question(non-blocking):** Just out of curiosity: Are the changes done here related to the camera? Why is the value copying required like so? What are the advantages?

src/core/nodeOps.ts
alvarosabu: @Tinoooo if you add several cameras in the template like: 

```ts
<TresPerspectiveCamera />
<TresOrthographic />
```

The last one to be registered is the active one. It's meant to be like this, if the `v-if` directive is used and it's false it will not call `insert`.  Here you can see how it works:

https://github.com/user-attachments/assets/11204706-4555-439d-ab3d-2a5b236f71c3

I just pushed the code to the playground so you can take a look http://localhost:5173/cameras

src/core/nodeOps.ts
alvarosabu: @Tinoooo there was a warning popping up for Orthographic camera when changing any prop mentioning that position on the Orthographic camera is readonly. Although I think this might be fixed by Cody's PR on main #1006

src/core/nodeOps.ts
Tinoooo: I'm aware of how the added code works right now. I just want to be sure that this is how it should work. When a user now adds a camera on runtime, be it through `v-if` or in another way, that exact camera will become active right away. The user has no way to prevent that. Imo this is too opinionated. We simply cannot know whether the added camera should be active right away or not.

Tinoooo: As mentioned above, I suggest omitting the line of code.

The example above would still work as you expect since at any given time there is just one camera in the scene. The computed `activeCamera` would take care of it.

alvarosabu: @Tinoooo can we tackle this conversation in the future 🙁? It was marked as a suggestion but is blocking the merging of the PR, I would like to move forward since this behavior is intended and this PR is blocking https://github.com/Tresjs/cientos/issues/615

Tinoooo: The playgrounds fail because the logic is missing that made sure that there is always an active camera. Omitting the `activeCameraUuid` and going back to the reordering method gives us that logic back. I created [a PR](https://github.com/Tresjs/tres/pull/1009) to show you what I mean.

Yes, we can tackle this in the future, but it seems a bit inefficient as we already put some thoughts into this.

alvarosabu: I just merged your PR into this one @Tinoooo. Let me know if there is something else pending

Tinoooo (approved):

PR #1006, opened 2025-05-01
fix(patchProp): harden props inference

CodyJasonBennett:
Hardens inference in `patchProp`, where unsafe property access was permitted before. This aligns with R3F v9.

This fixes edge cases like https://github.com/pmndrs/koota/issues/58, where inference can be fooled into breaking setters for non-atomic values.

### DIFF
--- a/src/core/nodeOps.ts
+++ b/src/core/nodeOps.ts
@@ -242,7 +242,7 @@ export const nodeOps: (context: TresContext) => RendererOptions<TresObject, Tres
   function patchProp(node: TresObject, prop: string, prevValue: any, nextValue: any) {
     if (!node) { return }
 
-    let root = node
+    let root: Record<string, unknown> = node
     let key = prop
 
     // NOTE: Update memoizedProps with the new value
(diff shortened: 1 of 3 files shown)


src/core/nodeOps.ts
```diff
@@ -312,11 +312,12 @@ export const nodeOps: (context: TresContext) => RendererOptions<TresObject, Tres
...
       // Refactor below to DRY.
-      const chain = key.split('-')
-      target = chain.reduce((acc, key) => acc[kebabToCamel(key)], root)
-      key = chain.pop() as string
-      finalKey = key
-      if (!target?.set) { root = chain.reduce((acc, key) => acc[kebabToCamel(key)], root) }
+      target = root
+      for (const part of key.split('-')) {
+        finalKey = key = kebabToCamel(part)
+        root = target as any
```
alvarosabu, line 318:
```suggestion
        root = target
```

You don't need to cast to any

alvarosabu (approved):
Hi @CodyJasonBennett thanks a lot for creating this PR, there is a small suggestion to remove the `as any` casting since its not needed and then we can merge. Thanks!

Issue #798, opened 2024-08-14
Support for kebab-case style component naming


Preyou:
### Description

I have been using the kebab-case style component with eslint constraints. However, Tres only supports UpperCamelCase, which breaks the style we already have, can we have support for kebab-case style component?

### Suggested solution

Do a naming conversion when reading the component

### Validations

- [X] I agree to follow this project's [Code of Conduct](https://github.com/Tresjs/tres/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/Tresjs/tres/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://tresjs.org/guide).
- [X] Check that there isn't [already an issue](https://github.com/tresjs/tres/issues) that reports the same bug to avoid creating a duplicate.

colinscz:
Hey @alvarosabu 
Could you please assign this issue to me? 
I would like to give it a go :)

alvarosabu:
@colinscz checkout from `next` 😜

colinscz:
This is now supported since Tres [5.1.0](https://github.com/Tresjs/tres/releases/tag/%40tresjs%2Fcore%405.1.0) and was addressed in #1193

PR #1015, opened 2025-05-24
feat: allow tres components written in kebab-case

colinscz:
Addresses #798

### DIFF
--- a/playground/vue/auto-imports.d.ts
+++ b/playground/vue/auto-imports.d.ts
@@ -66,6 +66,6 @@ declare global {
 // for type re-export
 declare global {
   // @ts-ignore
-  export type { Component, Slot, Slots, ComponentPublicInstance, ComputedRef, DirectiveBinding, ExtractDefaultPropTypes, ExtractPropTypes, ExtractPublicPropTypes, InjectionKey, PropType, Ref, MaybeRef, MaybeRefOrGetter, VNode, WritableComputedRef } from 'vue'
+  export type { Component, ComponentPublicInstance, ComputedRef, DirectiveBinding, ExtractDefaultPropTypes, ExtractPropTypes, ExtractPublicPropTypes, InjectionKey, PropType, Ref, MaybeRef, MaybeRefOrGetter, VNode, WritableComputedRef } from 'vue'
   import('vue')
 }
--- a/src/core/nodeOps.test.ts
+++ b/src/core/nodeOps.test.ts
@@ -136,6 +136,19 @@ describe('nodeOps', () => {
       expect(() => { nodeOps.createElement('TresBoxGeometry', undefined, undefined, null) }).not.toThrow()
       expect(() => { nodeOps.createElement('TresMeshNormalMaterial', undefined, undefined, null) }).not.toThrow()
     })
+
+    it('creates an instance with given tag when input is kebab-case', async () => {
+      // Setup
+      const tag = 'tres-mesh'
+      const props = { args: [] }
+
+      // Test
+      const instance = nodeOps.createElement(tag, undefined, undefined, props)
+
+      // Assert
+      expect(instance?.isObject3D).toBeTruthy()
+      expect(instance).toBeInstanceOf(Mesh)
+    })
   })
 
   describe('insert', () => {
--- a/src/core/nodeOps.ts
+++ b/src/core/nodeOps.ts
@@ -36,6 +36,13 @@ export const nodeOps: (context: TresContext) => RendererOptions<TresObject, Tres
     }
     if (tag === 'template') { return null }
     if (isHTMLTag(tag)) { return null }
+    if (tag.includes('-')) {
+      tag = kebabToCamel(tag)
+      const firstLetter = tag.charAt(0)
+      const upperFirstLetter = firstLetter.toUpperCase()
+      const restOfString = tag.slice(1)
+      tag = upperFirstLetter + restOfString
+    }
     let name = tag.replace('Tres', '')
     let obj: TresObject | null
 


playground/vue/auto-imports.d.ts
```diff
@@ -66,6 +66,6 @@ declare global {
 // for type re-export
 declare global {
   // @ts-ignore
-  export type { Component, Slot, Slots, ComponentPublicInstance, ComputedRef, DirectiveBinding, ExtractDefaultPropTypes, ExtractPropTypes, ExtractPublicPropTypes, InjectionKey, PropType, Ref, MaybeRef, MaybeRefOrGetter, VNode, WritableComputedRef } from 'vue'
+  export type { Component, ComponentPublicInstance, ComputedRef, DirectiveBinding, ExtractDefaultPropTypes, ExtractPropTypes, ExtractPublicPropTypes, InjectionKey, PropType, Ref, MaybeRef, MaybeRefOrGetter, VNode, WritableComputedRef } from 'vue'
```
colinscz: Was automatically adjusted 🤷🏽

src/index.ts
```diff
@@ -25,6 +25,7 @@ const plugin: TresPlugin = {
   install(app: App) {
     // Register core components
     app.component('TresCanvas', TresCanvas)
+    app.component('tres-canvas', TresCanvas)
```
colinscz: @alvarosabu not sure if this is the right approach. I am unsure about the TresCanvas part. 

The rest of the elements underneath seem to work fine after my tests with the playground and the unit tests. 

I commited the changes in the playground so it's easier to test with something right away.

src/core/nodeOps.ts
```diff
@@ -20,6 +20,12 @@ export const nodeOps: (context: TresContext) => RendererOptions<TresObject, Tres
     }
     if (tag === 'template') { return null }
     if (isHTMLTag(tag)) { return null }
+    // support kebab-case tags
+    if (tag.includes('-')) {
+      tag = tag.replace(/-([a-z])/g, (_, c) => c.toUpperCase())
+        .replace(/^[a-z]/, c => c.toUpperCase())
+      console.log('tag', tag)
```
colinscz: - [ ] TODO remove console.warn after final review has been done

playground/vue/src/pages/basic/Lights.vue
alvarosabu, line 1: Hey @colinscz I think it would be better if we create another route to test this and we leave the lights one as it was. You can create a new route under misc to test this

alvarosabu (changes requested):

playground/vue/src/pages/basic/Lights.vue
colinscz: Sure thing, I reused the playground scene for the kebab-case scenario to test it. But I am unsure regarding some changes. I commented the areas. Let me know if you want something else adjusted :)

src/core/nodeOps.test.ts
```diff
@@ -136,6 +136,34 @@ describe('nodeOps', () => {
...
+
+      // Test
+      const instance = nodeOps.createElement(tag, undefined, undefined, props)
+
+      // Assert
+      expect(instance?.isObject3D).toBeTruthy()
+      expect(instance).toBeInstanceOf(Mesh)
+    })
+
+    it('creates an instance with given tag using kebab-case notation and props ', async () => {
```
Tinoooo: **question:** Could these be non-async?
```suggestion
    it('creates an instance with given tag using kebab-case notation', () => {
      // Setup
      const tag = 'tres-mesh'
      const props = { args: [] }

      // Test
      const instance = nodeOps.createElement(tag, undefined, undefined, props)

      // Assert
      expect(instance?.isObject3D).toBeTruthy()
      expect(instance).toBeInstanceOf(Mesh)
    })

    it('creates an instance with given tag using kebab-case notation and props ', () => {
```

src/utils/template-compiler-options.ts
```diff
@@ -1,13 +1,14 @@
...
+  'tres-canvas',
   'TresLeches',
   'TresScene',
 ]
 
 const templateCompilerOptions = {
   template: {
     compilerOptions: {
-      isCustomElement: (tag: string) => (tag.startsWith('Tres') && !whitelist.includes(tag)) || tag === 'primitive',
+      isCustomElement: (tag: string) => ((tag.startsWith('Tres') || tag.startsWith('tres')) && !whitelist.includes(tag)) || tag === 'primitive',
```
Tinoooo: **suggestion(non-blocking):** I think this could be safer if we check for "Tres" + uppercase letter or "tres-". @alvarosabu Any objectives with changing this like shown below?

```suggestion
      isCustomElement: (tag: string) => ((/^Tres[A-Z]/.test(tag) || tag.startsWith('tres-')) && !whitelist.includes(tag)) || tag === 'primitive',
```

also applies for `vite.config.ts`

src/utils/template-compiler-options.ts
```diff
@@ -1,13 +1,14 @@
 const whitelist = [
   'TresCanvas',
+  'tres-canvas',
   'TresLeches',
   'TresScene',
```
Tinoooo, line 5: **question:** Should this be in kebab-case in the array too?

src/utils/template-compiler-options.ts
alvarosabu: @Tinoooo yes we should check both `TresGridHelper` and `tres-grid-helper`

src/utils/template-compiler-options.ts
colinscz: Hmm that is the question, I deliberately let TresLeches out of scope since I don't know if there are additional changes required in the specific package. And if I recall correctly I ran into an issue with `tres-scene` so that is why I did not adjust it. 

But it's been a couple days since I tried, let me fix the obvious things and then I try this out again in a separate commit.

src/utils/template-compiler-options.ts
colinscz: I adjusted both of them in 
4d95ad822330e7f312fe46b78f9e4c9efd5ebf33
4d95ad822330e7f312fe46b78f9e4c9efd5ebf33

src/index.ts
colinscz: after call with Tino: Try out without this line of code to see if it still works

src/utils/template-compiler-options.ts
Tinoooo: I guess TresScene can be omitted. @alvarosabu wdyt?

colinscz:
- [x]  Document possible usage of kebab-case component and option to deactivate the camelCase rule from the tresjs eslint config --> c1fe37c0e2d850946e8abf666f595a121ce1d5e5

src/index.ts
colinscz: Looks like it works without this registration too, I started the playground and it still worked. So I removed it @Tinoooo

docs/content/1.getting-started/1.index.md
```diff
@@ -53,3 +53,10 @@ You can write declarative Vue components:
 ```
 
 This approach **reduces the learning curve** significantly while preserving all the power and flexibility of Three.js.
+
+### Component Name Casing
+
+Throughout the guide, we are using PascalCase names when registering components. You can find more of the reasoning behind this in the official [Vue.js docs](https://vuejs.org/guide/components/registration.html#component-name-casing) This is because:
```
Tinoooo:
```suggestion
Throughout the guide, we are using PascalCase names when registering components. You can find more of the reasoning behind this in the official [Vue.js docs](https://vuejs.org/guide/components/registration.html#component-name-casing).

File: docs/content/1.getting-started/1.index.md [hunk 8]
```diff
@@ -53,3 +53,10 @@ You can write declarative Vue components:
...
+
+Luckily, Tres supports resolving kebab-case tags to components registered using PascalCase. This means a component registered as MyComponent can be referenced inside a Vue template (or inside an HTML element rendered by Vue) via both <MyComponent> and <my-component>. If you want to use the kebab-case notation in combination with the TresJS eslint config, you have to deactivate the following rule:
+`vue/component-name-in-template-casing` in your eslint configuration.
```
Tinoooo:
```suggestion
Luckily, Tres supports resolving kebab-case tags to components registered using PascalCase. This means a component registered as MyComponent can be referenced inside a Vue template (or inside an HTML element rendered by Vue) via both <MyComponent> and <my-component>. If you want to use the kebab-case notation in combination with the TresJS eslint config, you should deactivate the rule 
`vue/component-name-in-template-casing` in your eslint configuration.
```

src/utils/template-compiler-options.ts
Tinoooo: This thread is actually resolved. I'm just leaving it open for the question that arose above.

docs/content/1.getting-started/1.index.md
```diff
@@ -56,7 +56,11 @@ This approach **reduces the learning curve** significantly while preserving all
 
 ### Component Name Casing
 
-Throughout the guide, we are using PascalCase names when registering components. You can find more of the reasoning behind this in the official [Vue.js docs](https://vuejs.org/guide/components/registration.html#component-name-casing).
+Throughout the guide, we are using PascalCase names when registering components.
+::read-more
```
alvarosabu, line 60: @colinscz You can use the readmore by passing the link like:

```
::read-more{to="https://vuejs.org/guide/components/registration.html#component-name-casing"}
```

src/utils/template-compiler-options.ts
alvarosabu: Yes it can be ommited

PR #1017, opened 2025-05-28
feat: introduce useTres composable and update related documentation

alvarosabu:
- Added the `useTres` composable to provide a simplified API for accessing the TresJS context, including methods for `invalidate` and `advance`.
- Updated various components and documentation to utilize the new `useTres` composable, enhancing code clarity and maintainability.
- Included a new documentation page for `useTres` detailing its usage and properties.

### DIFF
--- a/docs/composables/use-tres.md
+++ b/docs/composables/use-tres.md
@@ -1,0 +1,64 @@
+# useTres
+
+The `useTres` composable provides access to a simplified version of the TresJS context. It is a wrapper around the `useTresContext` composable that provides a more convenient API for common use cases.
+
+## Usage
+
+```ts
+import { useTres } from '@tresjs/core'
+
+const { scene, renderer, camera, invalidate } = useTres()
+```
+
+:::warning
+`useTres` can be only be used inside a child component of a `TresCanvas` since this component acts as the provider for the context data.
+:::
+
+::: code-group
+
+```vue [App.vue]
+<script setup>
+import { TresCanvas } from '@tresjs/core'
+import SubComponent from './SubComponent.vue'
+</script>
+
+<template>
+  <TresCanvas
+  >
+    <SubComponent />
+  </TresCanvas>
+</template>
+```
+
+```vue [SubComponent.vue]
+<script lang="ts" setup>
+import { useTres } from '@tresjs/core'
+
+const { camera } = useTres()
+
+watchEffect(() => {
+  console.log(camera.value)
+})
+</script>
+```
+
+## Properties
+
+| Property | Description |
+| --- | --- |
+| **camera** | the currently active camera |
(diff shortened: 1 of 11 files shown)


docs/composables/use-tres.md
```diff
@@ -0,0 +1,63 @@
...
+| Property | Description |
+| --- | --- |
+| **camera** | the currently active camera |
+| **scene** | the [scene](https://threejs.org/docs/?q=sce#api/en/scenes/Scene) |
+| **renderer** | Contains the [WebGLRenderer](https://threejs.org/docs/#api/en/renderers/WebGLRenderer)  |
+| **sizes** | contains width, height and aspect ratio of your canvas |
+| **loop** | the renderer loop |
+| **controls** | the controls of your scene |
+| **raycaster** | the global raycaster used for pointer events |
+| **perf** | the performance monitor |
```
Tinoooo, line 55: **suggestion:** perf is not part of the context anymore. I suggest omitting this.

src/composables/useRenderer/useRendererManager.ts
```diff
@@ -310,12 +310,12 @@ export function useRendererManager(
 
   return {
     instance,
-
     isReady: readonly(isReady),
     advance,
     onRender,
     invalidate,
     canBeInvalidated,
+    amountOfFramesToRender,
```
Tinoooo, line 318: **question:** Are you sure this should be exposed? What is the use case? If it needs to be exposed: Shouldn't this be readonly?

src/composables/useTres/index.ts
```diff
@@ -0,0 +1,55 @@
...
+  return {
+    scene,
+    renderer: toValue(renderer.instance),
+    camera: camera.activeCamera,
+    sizes,
+    controls,
+    loop,
+    extend,
+    raycaster,
+    perf,
```
Tinoooo: **suggestion:** perf is not part of the context anymore. I suggest omitting this.

src/composables/useRenderer/useRendererManager.ts
alvarosabu: @Tinoooo, you guys use it in post-processing, so I guess yes?

https://github.com/Tresjs/post-processing/blob/7289c152b81ea9e386f7d4621e55fc61292b505f/src/core/three/EffectComposer.vue#L58-L66

Not including it would be a breaking change for you guys.

Tinoooo: Ah yes you are right. It still feels weird that this is exposed for some reason. Let's leave it in there for now. Maybe we have a better idea when useLoop is refactored. Maybe `render` from `useLoop` should provide a callback that provides a method that can be called to indicate that the frame was rendered. I'll keep that in the back of my head.

Issue #982, opened 2025-04-16
Refactor `useRenderer`


alvarosabu:
### Description

As a maintainer of TresJS, I would like to apply the following changes to the `useRenderer`:

- Drop renderer presets 
- Remove logic of renderer re-creation when TresCanvas props change (is not needed anymore since properties belonging to the renderer constructor args also affect canvas context, so recreation has no effect)
- Remove plainRenderer for defaults (Defaults should be explicit)
- Update TresCanvas types
- Add watchers for updatable renderer options at runtime

### WebGL options with set methods:

| Option | Set Method | Default Value |
|--------|------------|---------------|
| clearColor | `renderer.setClearColor` | #000000 |
| clearAlpha | `renderer.setClearAlpha` | 1 |

### WebGL options with direct setters

| Option | Set Method | Default Value | Comments  |
|--------|------------|---------------|---------------|
| shadows | `renderer.shadowMap.enabled` | false ||
| shadowMapType | `renderer.shadowMap.type` | PCFSoftShadowMap | Requires force material update |
| outputColorSpace | `renderer.outputColorSpace` | [SRGBColorSpace](https://threejs.org/docs/index.html#api/en/constants/Textures)|
| toneMapping | `renderer.toneMapping` | ACESFilmicToneMapping ||
| toneMappingExposure | `renderer.toneMappingExposure` | 1 ||
| dpr | `renderer.setPixelRatio` | 1 ||

### Additional context

Tres Canvas types:

```
export type WebGLRendererProps = TransformToMaybeRefOrGetter<Omit<WebGLRendererParameters, 'canvas'>>

// TODO move following type. Props shouldn't be defined by passing a ref.
export interface TresCanvasProps extends /* @vue-ignore */ WebGLRendererProps {
  /**
   * WebGL Context options (Readonly because they are passed to the renderer constructor)
   * They can't be changed after the renderer is created because they are passed to the canvas context
   */
  /**
   * Enables antialiasing, smoothing out edges of 3D objects.
   * Uses MSAA (Multisample Anti-Aliasing) when available.
   * @readonly
   * @default true (Opinionated default by TresJS)
   */
  antialias?: boolean

  /**
   * Enables stencil buffer with 8 bits.
   * Required for stencil-based operations like shadow volumes or post-processing effects.
   * @readonly
   * @default true
   */
  stencil?: boolean

  /**
   * Enables depth buffer with at least 16 bits.
   * Required for proper 3D rendering and depth testing.
   * @readonly
[...]

PR #1022, opened 2025-06-05
feat: 982 refactor userenderer

alvarosabu:


src/composables/useRenderer/useRendererManager.ts
```diff
@@ -116,30 +33,25 @@ export function useRendererManager(
...
-    premultipliedAlpha: toValue(options.premultipliedAlpha),
-    preserveDrawingBuffer: toValue(options.preserveDrawingBuffer),
-    logarithmicDepthBuffer: toValue(options.logarithmicDepthBuffer),
-    failIfMajorPerformanceCaveat: toValue(options.failIfMajorPerformanceCaveat),
   }))
 
-  const instance = shallowRef<WebGLRenderer>(new WebGLRenderer(webGLRendererConstructorParameters.value))
-
-  const amountOfFramesToRender = ref(0)
+  const frames = ref(0)
```
Tinoooo, line 204: **question:** What was the reason behind renaming this back? I remember that we intentionally named this `amountOfFramesToRender` to make it more readable and to make it express that it doesn't hold an array of frames.

src/composables/useRenderer/useRendererManager.ts
```diff
@@ -176,30 +88,22 @@ export function useRendererManager(
...
 
-  // since the properties set via the constructor can't be updated dynamically,
-  // the renderer is recreated once they change
-  watch(webGLRendererConstructorParameters, () => {
-    instance.value.dispose()
-    instance.value = new WebGLRenderer(webGLRendererConstructorParameters.value)
-
-    invalidateOnDemand()
-  })
-
```
Tinoooo, line 196: **question:** Is this not required anymore?

src/composables/useRenderer/useRendererManager.ts
```diff
@@ -209,28 +113,11 @@ export function useRendererManager(
     immediate: true,
   })
 
-  watch(() => options.clearColor, invalidateOnDemand)
-
   const { pixelRatio } = useDevicePixelRatio()
 
-  const getThreeRendererDefaults = () => {
```
Tinoooo, line 214: **question:** Don't we need this functionality anymore? This was here to make sure that the correct values are used when the props change to undefined. This way, we ensured that the outcome remains the same when equal property values are used.

src/composables/useRenderer/useRendererManager.ts
```diff
@@ -244,62 +131,69 @@ export function useRendererManager(
...
+  })
 
-      const valueInPreset = getValueFromPreset() as T
+  const watchersOptions = {
+    immediate: true,
+  }
 
-      if (valueInPreset !== undefined) { return valueInPreset }
+  // Watchers for updatable renderer options at runtime
+  watch(() => clearColorAndAlpha.value, (value) => {
```
Tinoooo: **issue:** The watchers do not trigger if `instance.value` changes. They did before as `watchEffect` was used.

src/composables/useRenderer/useRendererManager.ts
```diff
@@ -244,62 +131,69 @@ export function useRendererManager(
...
-      instance.value.setClearColor(
-        clearColor
-          ? normalizeColor(clearColor)
-          : new Color(0x000000), // default clear color is not easily/efficiently retrievable from three
-      )
+    if (isModeAlways.value) {
+      // If the render mode is 'always', ensure there's always a frame pending
+      frames.value = Math.max(1, frames.value)
     }
   })
```
Tinoooo: **question:** Doesn't the loop callback take care of that (line 97)?

src/composables/useRenderer/useRendererManager.ts
```diff
@@ -116,30 +33,25 @@ export function useRendererManager(
   {
     scene: Scene
     canvas: MaybeRef<HTMLCanvasElement>
-    options: UseRendererOptions
```
Tinoooo, line 119: **suggestion:** I suggest not using the `TresCanvasProps` here.
1. It does create a kind of circular (logical) dependency. `TresCanvas` uses this composable which uses a type from `TresCanvas.`
2. The elements in `options` should be optinally reactive (`MaybeRef`).

src/components/TresCanvas.vue
```diff
@@ -99,7 +80,7 @@ const slots = defineSlots<{
   default: () => any
 }>()
 
-const canvas = ref<HTMLCanvasElement>()
+const canvasRef = ref<HTMLCanvasElement>()
```
Tinoooo, line 68: **suggestion(non-blocking):** `useTemplateRef` could be used.

src/components/TresCanvas.vue
```diff
@@ -40,41 +39,23 @@ import { disposeObject3D, kebabToCamel } from '../utils/'
...
   useLegacyLights: undefined,
   preserveDrawingBuffer: undefined,
   logarithmicDepthBuffer: undefined,
   failIfMajorPerformanceCaveat: undefined,
   renderMode: 'always',
+  clearColor: '#000000',
+  clearAlpha: 1,
   enableProvideBridge: true,
+  toneMapping: ACESFilmicToneMapping,
+  shadowMapType: PCFSoftShadowMap,
```
Tinoooo, line 52: **suggestion:** I suggest using the defaults from three (`undefined`) unless we want to be opinionated here.

Tinoooo (changes requested):

src/components/TresCanvas.vue
alvarosabu: @Tinoooo I remember we discussed this in the past, we can't use templateRef, it was not working with the custom renderer

src/components/TresCanvas.vue
alvarosabu: This is done intentionally, the few that are opinionated are because most of production-quality scenes require those. The amount of code we reduced on the labs with this simple change was inmensive.

src/composables/useRenderer/useRendererManager.ts
alvarosabu: From https://github.com/Tresjs/tres/issues/982:

> Remove plainRenderer for defaults (Defaults should be explicit)

We agreed on this in a past meeting. We shouldn't be creating an extra empty renderer just to get the defaults. This adds overhead for resource disposal and its over-engineering, lets explicitly set the default ourselves with the types.

src/composables/useRenderer/useRendererManager.ts
alvarosabu: The truth is I'm still not comfortable with using long descriptive names when the context already gives the meaning (renderer). I was ok with `amountOfFramesToRender` but now that we need to expose it I think we are over-complicating It's called `frames` in every other custom renderer and in the context of the render the meaning is implicit @Tinoooo

src/composables/useRenderer/useRendererManager.ts
alvarosabu: From [the source ticket](https://github.com/Tresjs/tres/issues/982): 

> Remove logic of renderer re-creation when TresCanvas props change (is not needed anymore since properties belonging to the renderer constructor args also affect canvas context, so recreation has no effect)

We agreed on this on a past meeting

src/composables/useRenderer/useRendererManager.ts
alvarosabu: > The elements in options should be optionally reactive (MaybeRef).

Should they? this composable is used internally only, not open to public, with the context creation we are ensuring they all are defined and not reactive, @Tinoooo

src/composables/useRenderer/useRendererManager.ts
alvarosabu: Original

```
context.frames
```

vs

```
context.renderer.amountOfFramesToRender
```

Middle point

```
context.renderer.frames
```

Are there any other frames to be used on the context of the renderer so we do the differentiation?

src/composables/useRenderer/useRendererManager.ts
alvarosabu: > It does create a kind of circular (logical) dependency.

Why does create a circular dependency? We are passing props down to the composable, imagine now that the type for the props is in another file, wouldn't be exactly the same at the end?

```
import type { TresCanvasProps } from '@types`

Component props type -> props: TresCanvasProps -> useRenderer(props) 
```

src/composables/useRenderer/useRendererManager.ts
alvarosabu: The watchEffect for the setPixelRatio is up, L118

```
 watchEffect(() => {
    setPixelRatio(instance.value, pixelRatio.value, toValue(options.dpr))
  })
```

@Tinoooo do all props require having instance.value as dependency?

src/composables/useRenderer/useRendererManager.ts
Tinoooo: I prefer `amountOfFramesToRender` as it is more descriptive and tells the reader why this is here. But I'm fine with `frames` too.

src/composables/useRenderer/useRendererManager.ts
Tinoooo: Sorry, I remember now. 🫠

src/components/TresCanvas.vue
Tinoooo: Ah yes.

src/composables/useRenderer/useRendererManager.ts
Tinoooo: No. This is not a problem anymore since we are not recreating the renderer anymore.

src/composables/useRenderer/useRendererManager.ts
Tinoooo: ✅ I remember now.

src/composables/useRenderer/useRendererManager.ts
Tinoooo: Reason 2 is not valid anymore as we don't recreate the renderer anymore on argument changes. But I would still suggest defining the type of the options here and use it in `TresCanvas`. `useRendererManager` is the entity that knows what it needs to work so it is the source of the dependencies. `TresCanvas` only uses `useRenderManager` so it copies it's dependencies.
Now the options in the arguments of `useRendererManager` have values in their type that are not used in the arguments of the WebGLRenderer constructur (like `windowSize`, `enableProvideBridge`. `camera`). Somebody who adjusts `TresCanvasProps` wouldn't expect that the type is used that way.

src/composables/useRenderer/useRendererManager.ts
```diff
@@ -116,30 +33,25 @@ export function useRendererManager(
...
+    options: TresCanvasProps
     contextParts: Pick<TresContext, 'sizes' | 'camera' | 'loop'>
   },
 ) {
-  const webGLRendererConstructorParameters = computed<WebGLRendererParameters>(() => ({
-    alpha: toValue(options.alpha) ?? true,
-    depth: toValue(options.depth),
+  const instance = shallowRef<WebGLRenderer>(new WebGLRenderer({
```
Tinoooo: **suggestion:** Now that the renderer isn't recreated anymore, there is no need for this to be a `shallowRef`. It is written only once. I suggest using a simple variable.

src/composables/useRenderer/useRendererManager.ts
Tinoooo: This is removed now.

Tinoooo (approved):

PR #1026, opened 2025-06-09
feat!: 997 refactor events 2

Tinoooo:


Tinoooo:
**TODO:** update docs and migration guide. Inform about breaking changes concerning event listener names.

src/utils/pointerEvents.ts
```diff
@@ -0,0 +1,59 @@
+import type { Events } from 'vue'
+
+// Custom Tres events that aren't part of Vue's Events type
+type CustomTresEvents = 'onPointermissed' | 'onLostpointercapture'
+
+// All supported pointer events (Vue Events + custom Tres events)
+type SupportedPointerEvents = Extract<keyof Events, 'onClick'
```
Tinoooo, line 8: TODO: update docs and migration guide. Inform about breaking changes concerning event listener names,the camera prop (TresCanvas) and the drop of `blocks-pointer-events`.

src/composables/useLoop/index.ts
```diff
@@ -8,7 +8,7 @@ export function useLoop() {
     renderer,
     loop,
     controls,
-    events,
+    eventManager,
```
alvarosabu: **blocking:**

@Tinoooo we are not using the manager notation for either the renderer or camera, why do it for events? I would prefer if we keep it consistent as `events`

src/core/nodeOps.ts
```diff
@@ -254,7 +254,12 @@ export const nodeOps: (context: TresContext) => RendererOptions<TresObject, Tres
         // NOTE: Add node.__tres, if necessary.
         if (!node.__tres) { node = prepareTresInstance(node, {}, context) }
         node.__tres!.offPointerMissed?.()
-        node.__tres!.offPointerMissed = context.events?.onPointerMissed(nextValue).off
+        node.__tres!.offPointerMissed = context.events?.onPointerMissed(
```
alvarosabu: **Blocking:**
@Tinoooo I'm confirming with @bbohlender whether is intended to trigger once or not, the initial feedback is that:

> onPointerMissed is not standardized (made up by r3f not part of the w3c standard. The only use case I know that requires it, is for deselecting something, when the user clicks into the void

src/composables/useLoop/index.ts
Tinoooo: Okay. I changed the name back to `events`. It just sounds like an array. That's why I changed it.

src/core/nodeOps.ts
Tinoooo: I adjusted this so that `pointermissed` is emitted from `TresCanvas` whenever a click occurs and no interactive elements are hit.

PR #1029, opened 2025-06-13
feat: webgpu native support

### LINKED ISSUE #883: WebGPU support
### Description

WebGPU [Three.js r171](https://github.com/mrdoob/three.js/releases) has been released and it’s a major milestone for WebGPU integration with third-party libraries such as React Three Fiber and bundlers like Vite or Next.js. We can now do this with zero configuration:

Three.js Vanilla Three.js

```ts
import * as THREE from 'three'
import { WebGPURenderer } from 'three/webgpu'
 
const renderer = new WebGPURenderer()
```

R3F

```ts
import { Canvas } from '@react-three/fiber'
import { WebGPURenderer } from 'three/webgpu'
 
<Canvas gl={canvas => new WebGPURenderer({ canvas })}>
```

### Suggested solution

Enable a `renderer` prop to override the renderer, restricted to canvas-based WebGL/WebGPU renderers.

```vue
<script setup>
import { WebGPURenderer } from 'three/webgpu'
</script>

<template>
   <TresCanvas :renderer="WebGPURenderer" />
</template>
```

### Alternative

_No response_

### Additional context

_No response_

### Validations

alvarosabu:
# feat: WebGPU native support

This PR adds native WebGPU support to TresJS, allowing for better performance and modern rendering capabilities. The changes include:

## Key Changes
- Added WebGPU renderer support alongside existing WebGL renderer
- Improved renderer initialization with proper type handling
- Enhanced scene graph inspection capabilities for both WebGL and WebGPU contexts
- Added proper cleanup and context loss handling for both renderers

## Technical Details
- Modified renderer manager to handle both WebGL and WebGPU renderers
- Updated type definitions to support both renderer types
- Improved inspector tools to work with both rendering backends
- Added proper initialization for WebGPU context

## Usage

To enable WebGPU is as simple as: 

```ts
<script setup lang="ts">
const createWebGPURenderer = (ctx: TresRendererSetupContext) => {
  const renderer = new WebGPURenderer({
    canvas: toValue(ctx.canvas),
    // WebGPU specific configuration
    alpha: true,
    antialias: true,
  })
  return renderer
}
</script>

<template>
  <TresCanvas :renderer="createWebGPURenderer">
  </TresCanvas>
</template>
```

## Breaking Changes
None. The changes are backward compatible with existing WebGL implementations.

## Testing
- [ ] Tested with WebGL renderer
- [ ] Tested with WebGPU renderer
- [ ] Verified inspector tools work with both renderers
- [ ] Checked memory management and cleanup

## Related Issues
Closes #883

### DIFF
--- a/src/composables/useRenderer/useRendererManager.ts
+++ b/src/composables/useRenderer/useRendererManager.ts
@@ -9,9 +9,9 @@ import {
   useDevicePixelRatio,
 } from '@vueuse/core'
 import { Material, Mesh, WebGLRenderer } from 'three'
-import { computed, onUnmounted, readonly, ref, toValue, watch, watchEffect } from 'vue'
+import { computed, onUnmounted, ref, toValue, watch, watchEffect } from 'vue'
 import type { MaybeRef, ShallowRef } from 'vue'
-import { WebGPURenderer } from 'three/webgpu'
+import type { Renderer } from 'three/webgpu'
 
 // Solution taken from Thretle that actually support different versions https://github.com/threlte/threlte/blob/5fa541179460f0dadc7dc17ae5e6854d1689379e/packages/core/src/lib/lib/useRenderer.ts
 import { setPixelRatio } from '../../utils'
@@ -19,7 +19,8 @@ import { setPixelRatio } from '../../utils'
 import { logWarning } from '../../utils/logger'
 import type { SizesType } from '../useSizes'
 import type { UseCameraReturn } from '../useCamera'
-import type { TresScene } from 'src/types'
+import type { TresScene } from '../../types'
+import { isFunction, isObject } from '../../utils/is'
 
 /**
  * If set to 'on-demand', the scene will only be rendered when the current frame is invalidated
@@ -28,7 +29,7 @@ import type { TresScene } from 'src/types'
  */
 export type RenderMode = 'always' | 'on-demand' | 'manual'
 
-export type TresRenderer = WebGLRenderer | WebGPURenderer
+export type TresRenderer = WebGLRenderer | Renderer
 
 export interface RendererOptions {
   /**
@@ -206,23 +207,24 @@ export function useRendererManager(
     contextParts: { sizes, loop, camera },
   }: UseRendererOptions,
 ) {
-  let renderer: TresRenderer
+  const getRenderer = () => {
+    if (isFunction(options.renderer)) {
+      return options.renderer({
+        sizes,
+        scene,
+        camera,
+        loop,
+        canvas,
+      })
+    }
 
-  if (options.renderer && typeof options.renderer === 'function') {
-    renderer = options.renderer({
-      sizes,
-      scene,
-      camera,
-      loop,
-      canvas,
-    })
-  }
-  else {
-    renderer = new WebGLRenderer({
+    return new WebGLRenderer({
       ...options,
       canvas: unrefElement(canvas),
     })
   }
+
+  const renderer = getRenderer()
 
   const frames = ref(0)
   const maxFrames = 60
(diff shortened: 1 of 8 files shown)


src/composables/useTresContextProvider/index.ts
```diff
@@ -79,7 +79,7 @@ export function useTresContextProvider({
   ctx.loop.setReady(false)
   ctx.loop.start()
 
-  renderer.onReady(() => {
+  renderer.onReady?.(() => {
```
Tinoooo: **suggestion:** onReady always exists. I suggest omitting the optional chaining operator.
```suggestion
  renderer.onReady(() => {
```

src/composables/useTres/index.ts
```diff
@@ -36,8 +37,8 @@ export interface TresPartialContext extends Omit<TresContext, 'renderer' | 'came
   advance: () => void
 }
 
-export function useTres(): TresPartialContext {
-  const { scene, renderer, camera, sizes, controls, loop, extend, events } = useTresContext()
+export async function useTres(): Promise<TresPartialContext> {
+  const { scene, renderer, camera, sizes, controls, loop, extend, events } = await useTresContext()
```
Tinoooo: **suggestion:** useTresContext is not async. This should work without being async.
```suggestion
export function useTres(): TresPartialContext {
  const { scene, renderer, camera, sizes, controls, loop, extend, events } = useTresContext()
```

src/composables/useRenderer/useRendererManager.ts
```diff
@@ -8,12 +9,17 @@ import {
   useDevicePixelRatio,
 } from '@vueuse/core'
 import { Material, Mesh, WebGLRenderer } from 'three'
-import { computed, type MaybeRef, onUnmounted, type Reactive, readonly, ref, toValue, watch, watchEffect } from 'vue'
+import { computed, onUnmounted, readonly, ref, toValue, watch, watchEffect } from 'vue'
+import type { MaybeRef, ShallowRef } from 'vue'
+import { WebGPURenderer } from 'three/webgpu'
```
Tinoooo: **suggestion:** I suggest only importing the type and not use the instanceof check. I think this could improve the bundle size.

src/composables/useRenderer/useRendererManager.ts
```diff
@@ -22,6 +28,8 @@ import { logWarning } from '../../utils/logger'
  */
 export type RenderMode = 'always' | 'on-demand' | 'manual'
 
+export type TresRenderer = WebGLRenderer | WebGPURenderer
```
Tinoooo: **suggestion:** I suggest allowing any [Renderer](https://github.com/mrdoob/three.js/blob/4dd9c05c1a4f5f7e8ccfa9f8a5a338d70f3a1efc/src/renderers/common/Renderer.js#L46).

src/composables/useRenderer/useRendererManager.ts
```diff
@@ -242,11 +265,17 @@ export function useRendererManager(
 
   const isModeAlways = computed(() => toValue(options.renderMode) === 'always')
 
-  const renderEventHook = createEventHook<WebGLRenderer>()
+  const renderEventHook = createEventHook<TresRenderer>()
+
+  if (renderer instanceof WebGPURenderer) {
+    // Initialize the WebGPU context
+    renderer.init()
+    loop.setReady(true)
```
Tinoooo: **suggestion:** I suggest not calling this here and instead trigger the event hook. I think this is a little cleaner and makes the loop refactoring easier.

daLeiStrive:
现在支持webgpu了吗

Tinoooo:
> 现在支持webgpu了吗

It will be supported in TresJS version 5.

PR #1035, opened 2025-06-18
refactor: useLoop and useRenderLoop refactoring

Tinoooo:


src/composables/index.ts
```diff
@@ -1,11 +1,11 @@
 import UseLoader from './useLoader/component.vue'
 
 export * from './useCamera'
+export * from './useCreateRafLoop'
```
alvarosabu: @Tinoooo, I thought we were not going to expose it?

src/composables/useCreateRafLoop/index.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  let cycleFn: CycleFunction = defaultFunction
+
+  const eventHooks = {
+    before: createEventHook<RafLoopContext>(),
+    after: createEventHook<RafLoopContext>(),
+  }
+
+  const { pause, resume, isActive } = useRafFn(() => {
+    const getContextWithClock = (): RafLoopContext => ({
```
alvarosabu, line 26: Shouldn't we expose the clock?

src/composables/index.ts
Tinoooo: Yes, that's what we agreed on, but now that `useCreateRafLoop` is agnostic from the rendering and actually just a utility composable, I thought it might be handy for users too. It is still very specific, so I'm fine either way. Do you prefer to no expose it still?

alvarosabu: I see 🤔. On which occasions users would use instead of using `useLoop`? Don't you think it might confuse users?

src/composables/useCreateRafLoop/index.ts
Tinoooo: I thought about this. I'd suggest not expsing it as manually stopping or starting the clock ony would be unwanted behavior. I don't see any other use case than retrieving the `elapsed`, retrieving the `delta` and starting and stopping (bothlock and rafFn). All of these use cases are given. Am I overlooking something? I'm not totally against exposing it.

src/composables/index.ts
Tinoooo: They would use it to craeate their own loop basically. But I guess this is very very rare and if somebody wanted to, this is easily reproducable and will probably be adjusted according to the particular use case. So yes, this might just be confusing. Now that you asked this I'm more on the side of omitting the export. I have removed it now.

playground/vue/src/pages/basic/index.vue
```diff
@@ -57,15 +57,14 @@ const formattedShadowMapType = computed(() => {
...
+    <TresPerspectiveCamera :position="[5, 5, 5]" :look-at="[0, 0, 0]" />
     <OrbitControls />
-    <TresMesh :position="[0, 1, 0]" cast-shadow>
+    <TresMesh :position="[0, 0, 0]" cast-shadow>
       <TresBoxGeometry />
       <TresMeshStandardMaterial color="teal" :opacity="0.5" transparent />
     </TresMesh>
 
     <TresMesh
-      :rotation="[-Math.PI / 2, 0, 0]"
```
alvarosabu, line 68: Lol, why the change here? 😂 We lost the floor @Tinoooo 

Screenshot 2025-06-20 at 13 08 40

src/components/TresCanvas.vue
```diff
@@ -57,8 +57,11 @@ const props = withDefaults(defineProps<TresCanvasProps>(), {
 
 const emit = defineEmits<{
   ready: [context: TresContext]
-  render: [renderer: TresRenderer]
   pointermissed: [event: PointerEvent<MouseEvent>]
+  // TODO write explanation for differences
+  render: [renderer: TresRenderer] // TODO: should this be context? -> yes
+  beforeLoop: [context: TresContextWithClock]
+  loop: [context: TresContextWithClock]
```
alvarosabu, line 63: Screenshot 2025-06-20 at 14 40 39

For some reason I get any for the elapsed on the playground

playground/vue/src/pages/basic/index.vue
Tinoooo: The AI removed this and I didn't notice. You fixed it in your MR.

src/components/TresCanvas.vue
Tinoooo: This is not related to the referenced line. The parameter of `onLoop` in your screenshot is simply not typed.

alvarosabu: But should get the type from `TresContextWithClock` ?

alvarosabu (approved):

src/components/TresCanvas.vue
Tinoooo: Wdym?

PR #1044, opened 2025-06-21
refactor: 1043 render function triggers too often

Tinoooo:


src/composables/useLoop/index.ts
```diff
@@ -24,14 +24,11 @@ export const useLoop = () => {
...
 
-  const render = rendererManager.loop.replaceLoopFunction
-
   return {
     stop: rendererManager.loop.stop,
     start: rendererManager.loop.start,
     isActive: rendererManager.loop.isActive,
     onBeforeRender: eventHookBeforeRender.on,
     onRender: eventHookAfterRender.on,
-    render,
```
alvarosabu: **blocking**

@Tinoooo, this is another breaking change that will affect all users, :( . Playground for take over render mode is broken... 

Please lets keep the render callback here as it was before. Screenshot 2025-06-21 at 19 30 23

alvarosabu (changes requested):

src/composables/useLoop/index.ts
Tinoooo: I put the render function back in https://github.com/Tresjs/tres/pull/1044/commits/1d53b8aa9f6c22074ff703728b1789c8545c4099.
I also added a deprecation indicator. I'm also okay with omitting it if you want to keep the method there in future releases.

If you want we can also rename `replaceRenderFunction` to `setRenderFunction` if it suits you better.

alvarosabu (approved):
Thanks a lot for this PR and for tackling the discussions, I just tested it on post processing and all good, on demand rendering working as expected.

Issue #923, opened 2025-02-06
Error Boundary for TresCanvas


alvarosabu:
## Description

As a developer using TresJS, I need a way to gracefully handle and display runtime errors that occur within the TresCanvas component and its children. This would improve the development experience and make it easier to debug issues in production.

## Motivation
Currently, when an error occurs within TresCanvas or any of its child components (lights, meshes, materials, etc.) or threejs, the error can cause the entire application to crash or render a blank canvas without any meaningful feedback. This makes it difficult to:

1. Identify the source of the error
2. Debug issues in development
3. Provide a fallback UI in production
4. Handle errors gracefully without affecting the rest of the application

## Proposal

Add an Error Boundary feature to TresCanvas that would:

1. Catch runtime errors in:
   - Scene initialization
   - Component mounting/unmounting
   - Animation frame updates
   - Resource loading (textures, models, etc.)
   - Child component errors

2. Provide a customizable error display that shows:
   - Error message and stack trace (in development)
   - Component tree path to the error
   - Simplified error message (in production)

3. Include built-in error handling for common ThreeJS-specific errors:
   - WebGL context loss
   - Texture loading failures
   - Geometry errors
   - Material compilation errors

### Implementation Details

```vue
<template>
  <TresCanvas
    :error-boundary="true"
    :error-fallback="customErrorComponent"
    @error="handleError"
  >
    <!-- Scene content -->
  </TresCanvas>
</template>
```

The error boundary would inject necessary CSS styles automatically for the error display, eliminating the need for external CSS files. The styles would be scoped to the error boundary container to avoid conflicts.

### Default Error Display
The default error display would be a styled overlay with:
- Error message
- Stack trace (in development)
- Retry button
- Option to copy error details
- Responsive design that maintains the canvas aspect ratio

```typescript
// Internal styling injection
const errorStyles = `
  .tres-error-boundary {
    position: relative;
    width: 100%;
    height: 100%;
    background: rgba(30, 30, 30, 0.95);
    color: #fff;
    font-family: monospace;
    padding: 1rem;
    overflow: auto;
  }
  .tres-error-message {
    color: #ff5555;
    margin-bottom: 1rem;
  }
  .tres-error-stack {
    font-size: 0.9em;
    white-space: pre-wrap;
  }
  .tres-error-retry {
    background: #4a4a4a;
    border: none;
    color: white;
    padding: 0.5rem 1rem;
    margin-top: 1rem;
    cursor: pointer;
  }
`
```

[...]

PR #1048, opened 2025-06-22
feat: 1047 warn user if the canvas has no area

Tinoooo:


alvarosabu:
Should we scope this one as part of #923 ?

Tinoooo:
@alvarosabu 
> Should we scope this one as part of #923 ?

I understand #923 as something that catches console errors and warnings like the one I added. I'm not entirely sure what you mean with scoping it as part of #923. I suggest merging this MR to next and adjusting the warning if needed in #923.

src/components/TresCanvas.vue
```diff
@@ -244,6 +245,16 @@ onMounted(() => {
   if (import.meta.hot && context.value) { import.meta.hot.on('vite:afterUpdate', () => handleHMR(context.value as TresContext)) }
 })
 
+// warn if the canvas has no area
+onMounted(async () => {
+  await promiseTimeout(3000)
+
+  if (context.value && (!context.value.sizes.width || !context.value.sizes.height.value)) {
+    const windowSizePropName: keyof Pick<TresCanvasProps, 'windowSize'> = 'windowSize'
+    console.warn(`TresCanvas: The canvas has no area, so nothing can be rendered. Set it manually or use the prop ${windowSizePropName}.`)
```
alvarosabu:
```suggestion
    console.warn(`TresCanvas: The canvas has no area, so nothing can be rendered. Set it manually on the parent element or use the prop ${windowSizePropName}.`)
```

Issue #921, opened 2025-02-06
v5


alvarosabu:
### Description

This issue contains the discussion for the new major version v5.

## Goals 

- Stabilize ecosystem
- Improve the event handling system
- Reduce performance memory leaks
- WebGPU support
- Refactor several composables APIs

### Wishlist 

- useTresContextProvider orchestration separation of concerns into different composables
   - Rethink `useTresReady`
   - Move related logic to `useRenderer` https://github.com/Tresjs/tres/blob/90e1e94aac04681c4a428fcd7d139e444d33d789/src/composables/useTresContextProvider/index.ts#L112C2-L142C4
  - Move `setupDevtools` to TresCanvas
- Refactor `useTresContextProvider` createInjectionState https://vueuse.org/shared/createInjectionState/
- Refactor useRenderer
   - Don't recreate the renderer
   - Remove presets
   - Update renderer options that are updatable at runtime. (consider defaults without plainRenderer)
- Add `:renderer` prop. (WebGPU)
- Remove `useSeek`
- Remove spreading of useCamera composable on the context state
- Add CI for typecheck (vue-tsc)

alvarosabu:
Hi @AntoninRousset is not inside of our Roadmap right now due to capacity constraints, but we are open to having a community package similar of pmndr's package https://github.com/pmndrs/react-three-offscreen 

Would you be available to help us build that?

daLeiStrive:
So, when are you planning to support the features of WebGPU?

Tinoooo:
> So, when are you planning to support the features of WebGPU?

@daLeiStrive I answered your question before. WebGPU will be available with v5 (see [this issue](https://github.com/Tresjs/tres/issues/883)). See the sub-issues above.

PR #1060, opened 2025-07-23
chore: 1059 fix typecheck errors in docs

Tinoooo:


docs/app/components/examples/web-gpu/HologramCube.vue
```diff
@@ -45,8 +46,9 @@ material.colorNode = Fn(() => {
 })()
 
 watch(model, (newModel) => {
-  newModel.traverse((child) => {
-    if (child.isMesh) {
+  newModel?.traverse((child) => {
+    const isMesh = (child: Object3D): child is Mesh => 'isMesh' in child && !!(child.isMesh)
```
alvarosabu: (not-blocking):

I think I'm having a stroke with the readability of this line due to typescript 😵‍💫. Maybe we should add it as a util on the core package? @Tinoooo

Tinoooo: It exists in core, but unfortunately these utility methods are not exported. I didn't change that to keep the amount of changes small.
How would you change it to make it more readable? To me it looks like a pretty typical type guard.

alvarosabu: I think we should export it then, if users need to replicate this type guard to have type safety on traverse items of the model is not definitely not ideal, consider that most of our typescript users (included me) that type guard is not easy to read.

For end user we should aim for: 

```ts
if(isMesh(child)) {

}
```

Tinoooo: I created [this issue](https://github.com/Tresjs/tres/issues/921?issue=Tresjs%7Ctres%7C1062) for that. Solving it here would exceed the scope of this PR.
But still, imo the type guard would look similar inside in the core package as it does in the above mentioned code snippet. @alvarosabu Please let me know how you want it to look to be more readable.

alvarosabu (approved):

PR #1063, opened 2025-07-24
chore: 1061 fix eslint errors in docs

Tinoooo:


docs/package.json
```diff
@@ -28,6 +29,7 @@
   },
   "devDependencies": {
     "@nuxt/eslint": "^1.4.1",
+    "@tresjs/eslint-config": "^1.4.0",
```
alvarosabu: @Tinoooo the dependency is already available on the root package.json, you can remove it from here

docs/eslint.config.mjs
alvarosabu, line 1: @Tinoooo, why are you removing the nuxt lint?

docs/package.json
Tinoooo: I removed it.

docs/eslint.config.mjs
Tinoooo: I thought we would use our own config only. But now I saw that you added `@nuxt/eslint` to the modules. So I restored the nuxt config. Was adding the module intentional? I would suggest having only one source of truth for our eslint configs.

Tinoooo: We agreed in Discord to omit the nuxt module.

alvarosabu (approved):

PR #1067, opened 2025-07-29
feat: devtools communication strategy

alvarosabu:


alvarosabu:
@Tinoooo, I decided to remove the asset loading logic based on our recent conversation to find a better way to implement it (check #1080 ), and then I will also comment on the tab on the Nuxt module so we tackle it later. I will merge this one since its an internal thing and I need to close the v5 devtools for the release, we can iterate and improve it in the future

PR #1068, opened 2025-08-02
feat: make some of the utility methods available to users

Tinoooo:
This PR grew a little as I did some cleanup here and there.

Changes:

* exported type guard methods so they are available to users
* made use of the type guards throughout core
* removed unused methods
* moved utility methods out of `src/utils/index.ts`

Concerning the last point: I did not move all of the methods out of there to keep the changes of the PR smaller. But I suggest moving the remaining methods too as they are not general utilities but methods used in `nodeOps.ts`. I think it would be better to do this seperately. @alvarosabu Let's have a pair programming session to clean this up.


src/core/nodeOps.ts
```diff
@@ -18,7 +19,6 @@ export const nodeOps: (context: TresContext) => RendererOptions<TresObject, Tres
     if (!props.args) {
       props.args = []
     }
-    if (tag === 'template') { return null }
```
alvarosabu, line 21: comment:

I was worried about this removal, but I tested using templates in the code, and it works as expected. Can't remember why it was there in the first place 🤔 @Tinoooo

Tinoooo: @alvarosabu It is fine as the check one line below handles the template tag already. The line is redundant since temlate is a valid html tag.

PR #1085, opened 2025-09-16
chore: cookbook animations

alvarosabu:


docs/content/4.cookbook/2.basic-animations.md
```diff
@@ -0,0 +1,120 @@
...
+</template>
+```
+
+::read-more{to="/api/advanced/performance#reactivity-and-performance"}
+To read more about reactivity and performance in TresJS.
+::
+
+## Use `delta`
+
+The `onBeforeRender` callback provides a `delta` parameter, which represents the time elapsed since the last frame. This is useful for creating frame-rate independent animations.
```
colinscz:
```suggestion
The `onBeforeRender` callback provides a `delta` parameter, which represents the time elapsed since the last frame. This is useful for creating frame rate independent animations.
```

colinscz: I saw that  a couple lines down you write "frame rate", I don't mind which way, both works for me but it should be the same across the whole page :)

PR #1095, opened 2025-09-24
docs: Update Your First Scene

Baroshem:
Hey There!

This is my first (and hopefully one of many) contributions.

I was following the tutorial and I noticed that there is a small issue in the code sample - I believe the `:position` from the previous step was missing as without it the Donut is in the middle of the Grid helper while it should be above it to make it look the same as in the demo :)


alvarosabu (approved):

alvarosabu:
LGTM @Baroshem love to see you in Tres 💚

PR #1102, opened 2025-10-06
chore: dependency, scripts and file duplicate cleanup

Tinoooo:


alvarosabu (approved):
@Tinoooo thanks for fixing the netlify build.

PR #1186, opened 2025-10-07
fix(cientos):  scroll controls

JaimeTorrealba:
Currently, Scroll controls is broken, the fix is really simple

My plan is to update and improve this component, but I'll do it later in another PR for now let's just fix it


alvarosabu (approved):
Tested, LGTM @JaimeTorrealba glad to have you back

PR #1191, opened 2025-10-08
feat(cientos): support absolute-inspired positioning on `<ScreenSpace>`

lihbr:
This PR augments the existing Cientos `<ScreenSpace>` abstraction by adding support for CSS absolute-inspired positioning through new `top`, `bottom`, `left`, and `right` props.

This new behavior is supported for `PerspectiveCamera` and `OrthographicCamera`, takes into account the camera FOV and dynamic zoom updates.

This is especially useful when 3D experiences are here to "decorate" regular website content alongside the user's scroll for example. You can see a demo of it on this website for example: https://nuxt-starter-prismic-farbe.vercel.app

If you have the Netlify deploy preview for Cientos documentation, you should be able to preview the component in action.

Let me know if you have any comments/would like to make changes to the API/or abort this PR entirely 🙏

### DIFF
--- a/packages/cientos/src/core/abstractions/ScreenSpace.vue
+++ b/packages/cientos/src/core/abstractions/ScreenSpace.vue
@@ -39,8 +39,7 @@ const viewPlaneDimensions = computed(() => {
 
   const activeCamera = camera.activeCamera.value
   if (activeCamera instanceof PerspectiveCamera) {
-    const fov = activeCamera.fov
-    height = (2 * Math.tan(fov * Math.PI / 180 / 2) * props.depth) / zoom.value
+    height = (2 * Math.tan(activeCamera.fov * Math.PI / 180 / 2) * props.depth) / zoom.value
     width = height * sizes.aspectRatio.value
   }
   else if (activeCamera instanceof OrthographicCamera) {


JaimeTorrealba:
Hi @lihbr amazing having you here, I like the idea of this abstraction, but I'm not sure about the implementation, I find a couple of very confusing use cases. Would you like to discuss ideas on discord? And please next time open an issue before open a PR (specially with new ideas like this)

**First**: the "distance" prop in order to work properly needs to have the same distance as the camera (why not making automatic) other way it has some confusing behaviors. If you say in the docs (e.g. "top" prop works similar to CSS when absolute I would expect always to have the same value in all cases, not just when distance is same as the camera Z)

**Second**: Works erratically with Orthographic Camera (not a problem per se, but we need to define it on the docs) and work erratically with Perspective with some config for example when "zoom" is provided

**Third**: Even I like the idea of wrapping in a `Group` maybe (and just maybe) isn't that good because if I move my Mesh using position it gets quite confusing, for example:

```
    <AbsoluteGroup :distance="10" :left="0" :top="0">
      <TresMesh :position="[3,0,0]"> // here
        <TresTorusGeometry />
        <TresMeshNormalMaterial />
      </TresMesh>
    </AbsoluteGroup>

```
Maybe a composable/function that overwrite the position of the object¿?  `:position="Abs(x,y,z)"`?

**Lastly**:  And this is a possible improve. When moving the camera using controls, all the "benefit" of this abstraction get lost since isn't there anymore, so this one only sets the initial position. Could be handy to have a prop (we can of course discuss this one) to handle the case for always appears in that position. I said this because normally the camera in something that moves through the scenes.

Let me know what you think and happy to discuss this one :D

lihbr:
Hi @JaimeTorrealba, thanks for your feedback, and sorry for not opening an issue prior (it was a piece of code I was carrying around, and by discussing with Alvaro we thought it would be interesting to make it an abstraction for Cientos)

Happy to sync on Discord over ideas, I'm by far no Three.js expert, and I'm sure your insight on edge cases would be helpful! I like computing the camera distance automatically and the composable, especially 🙂

lihbr:
Hey @JaimeTorrealba,

I've made some updates following your feedback:
- Extracted the logic into a new `useAbsolutePosition` composable (turning `<AbsoluteGroup>` into a syntax sugar)[^1]
- Added support for orthographic camera and for `zoom`
- Made the group always display in front of the camera (essentially `absolute` to the camera render, as you suggested)

Regarding `distance`, I think keeping it a manually defined property actually make sense when you think of `<AbsoluteGroup>` as the 3D equivalent of CSS `position: absolute`, like a 3D `z-index` of sort, and that the original position of the object doesn't mean much in that scenario.

---

That being said, getting there made me realize `<AbsoluteGroup>` started to behave a lot like `<ScreenSpace>`

https://github.com/user-attachments/assets/e544a885-14f5-4329-a536-3c7c90f9c47f

So... I think instead of introducing a new abstraction, it might be better to do something like augmenting `<ScreenSpace>` with `top`, `bottom`, `left`, `right` props, to enable such usage:

```html
<!-- Keep the current behavior -->
<ScreenSpace :depth="5" />

<!-- Same as above, but anchored to the top left of the screen -->
<ScreenSpace :depth="5" :top="0" :left="0" />

<!-- Same as above, but anchored 30% off the top of the screen and 20% off the right -->
<ScreenSpace :depth="5" :top=".3" :right=".2" />

<!-- ... -->
```

Do you think that would make more sense? Happy to keep discussing and refactor in the agreed direction 🙂 

[^1]: I didn't update the doc for now, will wait until we're aligned on the interface if that's OK!

JaimeTorrealba:
I think you're right, I honestly think augmenting <ScreenSpace> instead of a new abstraction is the best way to go.

Question, what do you think is the best definition of the percentage value of the four new props (top, left, bottom, right)

Option 1: `:top="0.3"` as 30%
Option 2: `:top="30"` as 30%
Any other alternative? @lihbr

lihbr:
Great, I'll refactor in this direction :)

When we only have number types `0.3` screams more "percentage" to me than `30`, so I'd lean towards the former.

lihbr:
I believe it should be good for a proper review now, let me know if anything, happy to keep working on it!

Sorry for the back and forth on the API design 🙏

JaimeTorrealba (approved):
Nice and clean thanks for this @lihbr

PR #1200, opened 2025-10-11
docs: add info box to indicate community resources cover mostly Tres v4

colinscz:
Added an info box to the awesome resources, which tells viewers that most of the resources listed still cover v4 and that we will add new awesome resources along the way.


colinscz:
@alvarosabu the CI for some reason is not executed/in status pending on this PR, even after rebasing on main. Not sure if you can see more on your sight 🤔

PR #1206, opened 2025-10-16
chore: add dependency review job to check for vulnerable packages

colinscz:
This PR introduces the dependency review check for all future pull requests. 

For now I went with the default configuration but happy for input if we want to do some adjustments to the default policies.

Information about the action: 
The dependency review action scans your pull requests for dependency changes, and will raise an error if any vulnerabilities or invalid licenses are being introduced. The action is supported by an [API endpoint](https://docs.github.com/en/rest/dependency-graph/dependency-review?apiVersion=2022-11-28) that diffs the dependencies between any two revisions on your default branch.

https://github.com/actions/dependency-review-action

### DIFF
--- a/apps/cientos-docs-vitepress/.vitepress/theme/components/BackdropDemo.vue
+++ b/apps/cientos-docs-vitepress/.vitepress/theme/components/BackdropDemo.vue
@@ -52,6 +52,7 @@ const { hasFinishLoading, progress } = await useProgress()
           :rotation="[0, 0.5, 0]"
           :position="[0, 0.4, 0]"
           :scale="0.5"
+          draco
         />
       </Suspense>
       <Backdrop
--- a/packages/cientos/playground/vue/src/pages/staging/StageDemo.vue
+++ b/packages/cientos/playground/vue/src/pages/staging/StageDemo.vue
@@ -1,14 +1,21 @@
-<script setup lang="ts">
-import { Grid, OrbitControls, Stage } from '@tresjs/cientos'
-import type { LoaderProto } from '@tresjs/core'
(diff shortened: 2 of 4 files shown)


.github/workflows/dependency-review.yml
```diff
@@ -0,0 +1,24 @@
...
+  contents: read
+
+jobs:
+  dependency-review:
+    runs-on: ubuntu-latest
+    steps:
+      - name: 'Checkout Repository'
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+      - name: 'Dependency Review'
+        uses: actions/dependency-review-action@56339e523c0409420f6c2c9a2f4292bbb3c07dd3 # v4.8.0
```
Tinoooo: **question(non-blocking):** Could the version numbers be used here instead of the hashes?

colinscz: Yes indeed, I think the reason why for example Nuxt used the hash is because it's unique in case a version gets compromised, but I think since this one is out for some time I can change it to use the exact version number instead :)

alvarosabu: I would also prefer the version

colinscz: I added now the major version for the actions. Since both are official Github actions it should be fine, but if at some point we want to clarify or fix a version we can specify the minor or bugfix number: bb5d4688930ab1d139569bfa7ae315712211b970

colinscz:
@alvarosabu seems it also needs your approval as code owner :) I addressed the comment with the version.

PR #1209, opened 2025-10-23
feat: export type `TresCanvasProps` & `TresCanvasEmits` from TresCanvas.vue

Jungzl:
I want to import `TresCanvasProps` type but it's only available in `'@tresjs/core/dist/src/components/TresCanvas.vue'` with error
image

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -32,7 +32,7 @@ body:
     id: system-info
     attributes:
       label: System Info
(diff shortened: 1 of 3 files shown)


userquin:
Types should be fine now, there is no src folder at tresjs core dist folder with the new tsdown build at main released yesterday.

This can be closed.

Jungzl:
I don't think so

image


as well as `TresCanvasEmits`, they are not exported.

alvarosabu (approved):
Thanks @Jungzl 🙏🏻

PR #1221, opened 2025-11-18
feat(core): add unimport presets and unplugin vue components resolvers

userquin:
We should add some to the docs for these new subpackage exports (where chamo?).

I need to check `Tres*` components wrappers here https://github.com/Tresjs/tres/blob/main/packages/core/src/types/index.ts#L187-L191 🤔


alvarosabu:
> I need to check Tres* components wrappers here https://github.com/Tresjs/tres/blob/main/packages/core/src/types/index.ts#L187-L191 🤔

Hi @userquin the only real components to autoimport are the `TresCanvas`, the rest are part of of the custom renderer, they dont need explicit import since they are dinamically generated.

Does this make sense just for 1 component or it also autoimports composables?

alvarosabu:
Hey since this has been stale for a while I will go ahead and close it. Feel free to re-open if the questions asked can be addressed.

Issue #1223, opened 2025-11-19
[cientos]: not compatible with three v133~v166


Jungzl:
### Describe the bug

As I mentioned in https://github.com/Tresjs/tres/pull/1215#issuecomment-3483460913 & https://github.com/Tresjs/tres/pull/1215#issuecomment-3484167660

### Reproduction

https://stackblitz.com/edit/vitejs-vite-gsrhxaxk?file=src%2Fcomponents%2FFirstScene.vue

### Steps to reproduce

1. pnpm i
2. pnpm dev

### System Info

```shell
System:
    OS: Linux 5.0 undefined
    CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 20.19.1 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.8.2 - /usr/local/bin/npm
    pnpm: 8.15.6 - /usr/local/bin/pnpm
  npmPackages:
    @tresjs/cientos: ^5.1.1 => 5.1.1 
    @tresjs/core: ^5.1.1 => 5.1.1 
    three: 0.162.0 => 0.162.0 
    vite: ^7.2.2 => 7.2.2 
    vue: ^3.5.24 => 3.5.24
```

### Used Package Manager

pnpm

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/Tresjs/tres/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/Tresjs/.github/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://tresjs.org/guide).
- [x] Check that there isn't [already an issue](https://github.com/tresjs/tres/issues) that reports the same bug to avoid creating a duplicate.
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #1224, opened 2025-11-19
fix(cientos): improve WebGPU compatibility check in MeshReflectionMaterial

Jungzl:
fix: #1223

### DIFF
--- a/apps/docs/content/5.community/awesome-resources.md
+++ b/apps/docs/content/5.community/awesome-resources.md
@@ -13,6 +13,10 @@ description: A curated list of interesting TresJS community resources.
 You have a community plugin that helps solve a problem?
 Please open a pull request and let us know more details about what your plugin solves, so we can review it and add it to the list.
 
+::prose-note
+The following resources are mostly using Tres v4, we will add new resources covering v5 over time.
+::
+
 ## Recent talks on TresJS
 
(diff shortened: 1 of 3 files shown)


packages/cientos/src/core/materials/meshReflectionMaterial/index.vue
```diff
@@ -282,7 +281,9 @@ const { onBeforeRender } = useLoop()
 onBeforeRender(({ renderer, scene, camera }) => {
   const parent = (materialRef.value as any)?.__tres?.parent
   if (!parent) { return }
-  if (renderer instanceof WebGPURenderer) {
+  // WebGPU compatibility: avoid importing three/webgpu (not exported in older versions).
+  // Detect WebGPU renderer by its flag property defined in its class (`isWebGPURenderer: true`).
+  if ((renderer as any)?.isWebGPURenderer === true) {
```
JaimeTorrealba: Blocker: 
   Can we use a type guard here so we avoid casting to `any`

packages/cientos/src/core/materials/meshReflectionMaterial/index.vue
```diff
@@ -282,7 +281,9 @@ const { onBeforeRender } = useLoop()
 onBeforeRender(({ renderer, scene, camera }) => {
   const parent = (materialRef.value as any)?.__tres?.parent
```
JaimeTorrealba: Non-blocker:
   we should also avoid any here, but we'll take care of it

JaimeTorrealba (changes requested):

PR #1225, opened 2025-11-19
chore: remove redundant spaces from the envinfo command args

Jungzl:
To print three and vue version info correctly


alvarosabu:
Thanks @Jungzl

PR #1229, opened 2025-11-26
fix(core): import version directly from package.json for data-tres attribute

Jungzl:
to reduce bundle size

### DIFF
--- a/packages/cientos/playground/vue/src/pages/abstractions/ImageDemo.vue
+++ b/packages/cientos/playground/vue/src/pages/abstractions/ImageDemo.vue
@@ -3,7 +3,6 @@ import { TresCanvas } from '@tresjs/core'
 import { Image, OrbitControls } from '@tresjs/cientos'
 import { Color, DoubleSide, FrontSide, NoToneMapping } from 'three'
 import { TresLeches, useControls } from '@tresjs/leches'
-import '@tresjs/leches/styles'
 
 const URLS = [
   'https://upload.wikimedia.org/wikipedia/commons/1/13/20220713-great-tit.jpg',
--- a/packages/cientos/src/core/abstractions/Image/component.vue
+++ b/packages/cientos/src/core/abstractions/Image/component.vue
@@ -48,19 +48,22 @@ export type ImageProps = {
    * THREE.Side of the image material. [See THREE.material.side](https://threejs.org/docs/?q=material#api/en/materials/Material.side)
    */
   side?: Side
-} & ({
-  /**
-   * Image texture to display on the geometry.
-   */
-  texture: Texture
-  url?: never
-} | {
-  texture?: never
-  /**
-   * Image URL to load and display on the geometry.
-   */
-  url: string
-})
+} & (
+  | {
+    /**
+       * Image texture to display on the geometry.
+       */
+    texture: Texture
+    url?: never
+  }
+  | {
+    texture?: never
+    /**
+       * Image URL to load and display on the geometry.
+       */
+    url: string
+  }
+)
 
 const props = withDefaults(defineProps<ImageProps>(), {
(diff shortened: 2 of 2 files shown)


alvarosabu:
@Jungzl out of curiosity, which version of node do you have? We noticed issues with node 24, tsdown/rolldown and attribute importing for json files.

Jungzl:
@alvarosabu I use node22.15.1 which is configured in .nvmrc, but yes node24 will throw error

Jungzl:
image

build successfully with node22&24 @alvarosabu

userquin:
@Jungzl  `dist/tres.js` just with the package version?

userquin:
weird, your screenshot logs using same node v22

Jungzl:
@userquin yes
image

the first one is v22.15.1 but after `fnm use lts-jod` it uses v22.21.1 and the second uses v24.11.1
image

userquin (approved):
LGTM

@alvarosabu maybe we should add something about this to the docs (supported node versions to build the repo/packages) adding also the sqlite issue with nuxt (force us to remove all node_modules to just rebuild sqlite when switching between node versions)

Jungzl:
@userquin 
I always define latest lts version(e.g. lts-krypton, which is v24.11.1 right now) in .nvmrc for developing and supports at least two latest major lts versions(e.g. lts-krypton & lts-jod) in CI for compatibility.

I assume all contributors have installed tools like nvm, fnm or volta that can manage Node versions.

userquin:
I use nvm, at elk.zone we use lts/*:
https://github.com/elk-zone/elk/blob/main/.nvmrc

userquin:
@Jungzl @alvarosabu something still wrong when releasing the package 🤔 , this is the latest `dist/tres.js` 5.1.1 released a few minutes ago https://registry.npmjs.org/@tresjs/core/-/core-5.1.1.tgz:

image

Jungzl:
@alvarosabu  
image
v5.1.1 was released 22days ago.

lastest version should be v5.2.0 but something went wrong in CI(changelogs updated though).
might be some OIDC issues

userquin:
@Jungzl so the latest release  is wrong? the npm registry seems to have 5.1.1 as latest, I just check it via https://arethetypeswrong.github.io/ adding `@tresjs/core` in the input => resolving to 5.1.1 weird, checking npm registry (it is wrong https://www.npmjs.com/package/@tresjs/core), release seems to be fine here at GH (also wrong ?) => maybe latest release not changed here:

image

userquin:
Latest publish is wrong, maybe NX issue? Check https://github.com/Tresjs/tres/actions/runs/19477064478/job/55739451933, expand `Publish packages` entry:

image

Jungzl:
I don't think they were ever published. The Publish action has not run recently. The action you mentioned happened three weeks ago. @userquin

userquin:
then, why do we get GH notification? I'm missing something how it is being published and released, there is no publish workflow https://github.com/Tresjs/tres/actions/workflows/publish.yaml (I use bumpp from Anthony Fu + reusable workflow from Kevin Deng https://github.com/vite-pwa/vite-plugin-pwa/blob/main/.github/workflows/release.yml or in a monorepo https://github.com/userquin/unvuetify-monorepo/blob/main/.github/workflows/release.yml), maybe wrong `contents: read` here?

alvarosabu:
Hi @userquin @Jungzl I ran the manual publishing action here https://github.com/Tresjs/tres/actions/runs/20101821756/job/57674526626, the core was the only published correctly due the trusted publishers workflow that we are changing right now. 

I think your confusion @userquin is that there is a 2 steps release process, the bump is done locally using `nx release --skip-publish` and then the release it self is done in a manual action here in github

PR #1230, opened 2025-11-27
fix(core): reuse type `TresPointerEvent`

Jungzl:
When I use `TresCanvasEmits` in my custom component, `vue-tsc` does not generate correct dts may due to complex types.
Before:
image
After:
image


alvarosabu (approved):
Thanks a lot @Jungzl

Issue #1228, opened 2025-11-26
[Vue warn]: provide() can only be used inside setup().


Jungzl:
### Describe the bug

vue@3.5.25 bring a warning about the usage of `provide()` though it doesn't cause bugs rn.
ref: https://github.com/vuejs/core/pull/13954

I think it is the `useTresContextProvider` inside `onMounted`
https://github.com/Tresjs/tres/blob/4877a92b84bfaa52cac6060986ecf62fca50650a/packages/core/src/components/TresCanvas.vue#L153-L161

### Reproduction

https://stackblitz.com/edit/vitejs-vite-uay94k5d?file=src%2FApp.vue

### Steps to reproduce

open devtools to see the warning

### System Info

```shell
System:
    OS: Windows 11 10.0.26200
    CPU: (24) x64 13th Gen Intel(R) Core(TM) i7-13700
    Memory: 820.05 MB / 15.77 GB
  Binaries:
    Node: 24.11.1
    Yarn: 1.22.22
    npm: 11.6.2
    pnpm: 10.23.0
    bun: 1.3.3
  Browsers:
    Chrome: 142.0.7444.176
    Edge: Chromium (140.0.3485.54)
    Firefox: 145.0.1
    Internet Explorer: 11.0.26100.1882
  npmPackages:
    @tresjs/cientos: ^5.1.1 => 5.1.1 
    @tresjs/core: ^5.1.1 => 5.1.1 
    rolldown-vite:  7.2.5 
    three: ^0.181.2 => 0.181.2 
    vue: ^3.5.25 => 3.5.25
```

### Used Package Manager

pnpm

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/Tresjs/tres/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/Tresjs/.github/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://tresjs.org/guide).
- [x] Check that there isn't [already an issue](https://github.com/tresjs/tres/issues) that reports the same bug to avoid creating a duplicate.
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #1232, opened 2025-11-28
refactor(core): extract canvas context logic to Context.vue and composables

Jungzl:
follows up: #1230 
fix: #1228


packages/core/src/components/TresCanvas.vue
```diff
@@ -297,5 +59,32 @@ export type TresCanvasEmits = {
...
       ...$attrs.style as Object,
     }"
-  ></canvas>
+  >
+    <Context
+      v-if="canvasRef"
+      ref="contextRef"
+      :canvas="canvasRef"
+      v-bind="props"
+      @ready="emit('ready', $event)"
```
Tinoooo, line 105: **suggestion(non-blocking):** This should be more type-safe. If we add a new event to `TresCanvasEmits`, Typescript should complain. I think the best way to achieve this in a nice clean way is using something like [useForwardPropsEmits](https://www.radix-vue.com/utilities/use-forward-props-emits.html#useforwardpropsemits). I have the code ready, so I suggest merging this PR first so I can add the code afterwards.

packages/core/src/components/types.ts
Tinoooo, line 1: **suggestion:** We generally want to keep the types in the component files. I suggest moving this back to the component.

Jungzl: No problem.

packages/core/src/components/TresCanvas.vue
Jungzl: Thank you for your suggestion and let me know there's a utility function like `useForwardPropsEmits`. My implementation was indeed not elegant.

packages/core/src/components/types.ts
Tinoooo: Thank you for the quick fix.

packages/core/src/components/TresCanvas.vue
Tinoooo: My plan is to introduce the one from Reka UI by copying it. I think this composable is a good candidate for VueUse. I created a [feature request](https://github.com/vueuse/vueuse/issues/5202).

Jungzl: Yeah, that util looks great and well-typed. Could you help me review #1233 as well before you make a PR for this?

Jungzl: I don't know why `pkg.pr.new` not running per PR and I really need that feature that I have to patch or link local files to develop.

Jungzl: But honestly, introducing a few kilobytes of dependencies for type safety requires some trade-offs. After all, adding a new event to `TresCanvasEmits` might not be that likely or frequent.

PR #1233, opened 2025-11-28
feat(core): extract context management logic to useTresContextManager composable

Jungzl:
follows up: #1232 
close: #1210 

like [`createRoot`](https://github.com/pmndrs/react-three-fiber/blob/2541e81fb6ddc22d0869b9eb5cdbedcbbc62324c/packages/fiber/src/core/renderer.tsx#L142) from `@react-three/fiber` or [`createThrelteContext`](https://github.com/threlte/threlte/blob/0f4563474af18f3a073b60bada7d8edf608a499a/packages/core/src/lib/context/createThrelteContext.svelte.ts#L23) from `@threlte/core`

finally, I can easily integrate `tresjs` in map app.

image

### DIFF
--- a/packages/core/src/composables/useTresContextManager/index.ts
+++ b/packages/core/src/composables/useTresContextManager/index.ts
@@ -75,7 +75,7 @@ export function useTresContextManager(
 
   const mountCustomRenderer = (context: TresContext, empty = false) => {
     const InternalComponent = createInternalComponent(context, empty)
-    const { render } = createRenderer(nodeOps(context))
+    const { render } = createRenderer(nodeOps({ context, options: props.customRendererOptions }))
     render(h(InternalComponent), scene.value as unknown as TresObject)
   }
 


packages/core/src/composables/useTresContextManager/index.ts
```diff
@@ -0,0 +1,193 @@
...
+export function useTresContextManager(
+  scene: ShallowRef<TresScene>,
+  context: ShallowRef<TresContext>,
+  props: TresCanvasProps,
+  emit: ReturnType<typeof defineEmits>,
+  slots: ReturnType<typeof defineSlots>,
+) {
+  const instance = getCurrentInstance()
+
+  const createInternalComponent = (context: TresContext, empty = false) =>
```
Tinoooo, line 35: **issue:** Moving all this from the TresCanvas component to this composable is not desirable. The composables role is to initialize and orchestrate all the things around three.js while showing the dependencies between the respective elements (like renderer needs camera, event manager needs the renderer,...). The things that are moved here are very specific to the TresCanvas. Moving them harms the architecture and the readability imo. @Jungzl Maybe you can explain a bit what kind of fix or feature the change brings for you.

Jungzl: Thanks for your feedback. 

The motivation behind this PR is to enable rendering three.js content inside a host-provided WebGL layer (e.g., a map’s WebGL [layer](https://docs.mapbox.com/mapbox-gl-js/example/add-3d-model/)) while reusing Tres’s scene/camera lifecycle, scheduling, and state management—without duplicating the internal logic from the Context component. 

But Tres doesn’t expose a low-level primitive equivalent to `createRoot` (from R3F) or `createThrelteContext` (from Threlte). If we attempt to integrate directly by mounting into an external canvas/context, we end up replicating much of Context.vue’s orchestration. To avoid that, this PR extracts and exports the logic as a composable (`useTresContextManager`), paired with `useTresContextProvider`.

The code looks like below:
```vue
<!--  CustomTresLayer.vue -->
<script setup lang="ts">

//...

const slots = defineSlots<{
  default: () => any
}>()

extend(THREE)

const createRenderer: RendererOptions['renderer'] = ctx => {
  const renderer = new WebGLRenderer({
    ...ctx,
    canvas: hostCanvas,
    context: hostGLContext,
  })
  renderer.autoClear = false
  return renderer as TresRenderer
}

const scene = shallowRef(new Scene())
const context = shallowRef(
  useTresContextProvider({
    scene: scene.value,
    canvas: hostCanvas,
    rendererOptions: { ...props, renderer: createRenderer }
  })
)

const { dispose } = useTresContextManager(scene, context, props, emit, slots)

// Bridge render requests to host
watchEffect(() => {
  context.value?.renderer?.replaceRenderFunction(notifySuccess => {
    host.render()  // Trigger host redraw
    notifySuccess()
  })
})

// Sync camera to host view in render callback
const render = () => {
  const { camera, scene } = context.value
  const activeCamera = camera.activeCamera.value
  
  // Apply host camera parameters
  const { near, far, fov, position, lookAt } = host.getCameraParams()
  activeCamera.near = near
  activeCamera.far = far
  activeCamera.fov = fov
  activeCamera.position.set(...position)
  activeCamera.lookAt(...lookAt)
  activeCamera.updateProjectionMatrix()
  
  // Render with host state
  renderer.render(scene.value, activeCamera)
  renderer.resetState()
}
</script>

<template>
  <CustomLayer :render="render" />
</template>
```

PR #1234, opened 2025-12-01
perf(cientos): improve tree-shake for image and outline materials

init-qy:
aslo remove `export default`


packages/cientos/src/core/abstractions/Image/ImageMaterialImpl.ts
```diff
@@ -5,7 +5,7 @@ import { Color, Vector2 } from 'three'
  * NOTE: Source:
  * https://threejs.org/docs/?q=material#api/en/materials/Material.transparent
  */
-const imageMaterialImpl = shaderMaterial(
+export const imageMaterialImpl = /* @__PURE__ */ shaderMaterial(
```
JaimeTorrealba: I think just a small change here, can this be export as PascalCase like the others instead of camelCase, please.

`ImageMaterialImpl` with uppercase at the beginning

JaimeTorrealba (approved):
Thanks for this one

PR #1241, opened 2025-12-09
chore(ci): update Node.js version in .nvmrc and trusted publishers setup

alvarosabu:


.github/workflows/publish.yaml
```diff
@@ -4,12 +4,12 @@ on:
...
 
 jobs:
-  test:
+  publish:
     name: Publish
     runs-on: ubuntu-latest
     permissions:
       contents: read
-      id-token: write # needed for provenance data generation
+      id-token: write # Required for OIDC trusted publishing
```
colinscz, line 12: You might have to change the contents permission to write if the tag and release creation here on Github is done in this step to or if you want to automate this here.

.github/workflows/publish.yaml
```diff
@@ -20,8 +20,6 @@ jobs:
 
       - uses: ./.github/actions/setup-node
 
-      - run: pnpm config set //registry.npmjs.org/:_authToken ${{ secrets.NPM_TOKEN }}
-
       - name: Build publishable packages
```
colinscz, line 23: ℹ️  I am unsure if the provenance flag is indeed needed here: https://pnpm.io/cli/publish#--provenance 

Let's try it without it, I checked on some other repositories and they don't seem to use the flag. So it should work without it.

colinscz (approved):

PR #1262, opened 2025-12-23
fix(core): add prefix option for primitives

alvarosabu:
### Summary
  Adds a new `options` prop to `<TresCanvas>` with a `primitivePrefix` configuration that allows users to customize the primitive component name.

  ### Changes
  - **Core**: Added `TresCustomRendererOptions` interface with `primitivePrefix` option
  - **TresCanvas**: New `options` prop for custom renderer configuration
  - **Context**: Pass options through to the custom renderer's nodeOps
  - **NodeOps**: Check for prefixed primitive tag (e.g., `myprimitive` when prefix is `"my"`)
  - **Types**: Fixed module augmentation (removed incorrect `export` keyword in `GlobalComponents`)
  - **Docs**: Added documentation to TresCanvas API and Primitives advanced guide
  - **Playground**: Added example demonstrating the feature with TypeScript support

  ### Usage
  ```vue
  <script setup lang="ts">
  const options = {
    primitivePrefix: 'my', // Use <myprimitive> instead of <primitive>
  }
  </script>

  <template>
    <TresCanvas :options="options">
      <myprimitive :object="threeObject" />
    </TresCanvas>
  </template>
```
  
### TypeScript Support

  Users can extend Vue's global components for full intellisense:
```
  // types/tres.d.ts
  import type { TresPrimitive } from '@tresjs/core'
  import type { DefineComponent } from 'vue'

  declare module 'vue' {
    interface GlobalComponents {
      myprimitive: DefineComponent<TresPrimitive>
    }
  }

  export {}
```

###  Benefits

  - Avoid naming conflicts with other libraries
  - Support custom naming conventions
  - Maintain full TypeScript intellisense support

TRES-186

### DIFF
--- a/packages/core/src/components/Context.vue
+++ b/packages/core/src/components/Context.vue
@@ -1,7 +1,7 @@
 <script setup lang="ts">
 import { PerspectiveCamera, Scene, WebGLRenderer } from 'three'
 import type { App } from 'vue'
-import type { TresObject, TresScene } from '../types'
+import type { TresCamera, TresContextWithClock, TresObject, TresPointerEvent, TresScene } from '../types'
 import * as THREE from 'three'
 import {
   createRenderer,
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: apps/docs/content/3.api/1.components/tres-canvas.md [hunk 1]
```diff
@@ -174,6 +174,36 @@ For detailed technical information about prop reactivity, see [GitHub Issue #982
   ::::field{name="windowSize" type="boolean"}
   **⚡ Reactive** - Whether the canvas should be sized to the window. When true, canvas will be fixed positioned and full viewport size.
   ::::
+
+  ::::field{name="options" type="TresCustomRendererOptions"}
+  **⚡ Reactive** - Configuration options for the TresJS custom renderer:
```
Tinoooo: **issue:** This says `Reactive`, but I was not able to reactively set the `primitivePrefix`. The scene stopped working when doing so. Imo this is nothing that needs to be reactive as it is nothing that is usually set on runtime.

packages/core/src/components/Context.vue
```diff
@@ -111,6 +111,7 @@ const context = shallowRef<TresContext>(useTresContextProvider({
   canvas: props.canvas,
   windowSize: props.windowSize ?? false,
   rendererOptions: props,
+  options: props.options ?? { primitivePrefix: '' },
```
Tinoooo: **suggestion:** I suggest naming this something more precise. customRendererOptions for example.

packages/core/src/composables/useTresContextProvider/index.ts
```diff
@@ -21,6 +25,7 @@ export interface TresContext {
   controls: Ref<TresControl | null>
   renderer: UseRendererManagerReturn
   events: ReturnType<typeof useEventManager>
+  options: TresCustomRendererOptions
```
Tinoooo: **suggestion:** I suggest not piping this through the composable. The reasons are:

1. I understand `useTresContextProvider` as the entity that initializes and orchestrates everything three.js–related. Custom renderer logic feels misplaced here.
2. The primitive prefix does not work reactively (and shouldn’t, in my opinion), so it does not belong in a composable.
3. The context can be accessed in subcomponents of the Canvas. Having the primitive prefix there doesn’t provide any benefit, in my opinion. It would also allow setting the primitive prefix from there, which does not work and feels architecturally undesirable.

Tinoooo (changes requested):
I created #1265 to show how I would solve the review threads I created.

alvarosabu:
@Tinoooo can you help me solve the conflicts with the typescript PR?

Tinoooo:
> @Tinoooo can you help me solve the conflicts with the typescript PR?

done ✅

PR #1263, opened 2025-12-24
chore(core): make use of pixelRatio

Tinoooo:
follow-up for #1261: Let's use the newly exposed `pixelRatio` in the renderer.


alvarosabu (approved):
Nice catch LGTM

PR #1265, opened 2025-12-27
fix(core): add prefix option for primitives

Tinoooo:
These are the changes I propose concerning my review for #1262 plus one change in pnpm-workspace.yaml to make the build work (benefiting from https://github.com/sxzz/rolldown-plugin-dts/pull/152).


Tinoooo:
> Hi @Tinoooo thanks for improving the implementation, yours is way more elegant.
> 
> Should we merge it on the other PR or directly on main?

I'm glad you like it. The merging order doesn't really matter. I just directed thir PR to #1262 to visualize the changes better.

PR #1268, opened 2026-01-13
fix(docs): fix OrbitControls example to use camera instead of renderer

moliyu:
as camera is a ref and maybe undefined which will result page load err


alvarosabu (approved):
Hey @moliyu thanks a lot

PR #1275, opened 2026-01-19
fix: tres webgpu broken on three v181

alvarosabu:
### Summary

- Updates Three.js and `@types/three` to version 0.181.0
- Fixes renderer initialization to properly support WebGPU's async `init()` requirement
- Fixes camera registration timing to ensure user cameras take precedence over the default camera

### Problem

Three.js v0.181.0 introduced a breaking change where WebGPU renderers require `await renderer.init()` before any operations can be performed. Additionally, when using `@tresjs/post-processing`, effects like `Pixelation` would capture a stale camera reference because:

1. The default camera was registered first and became `activeCamera`
2. User cameras (e.g., `<TresPerspectiveCamera>`) were registered via `nodeOps.insert()` without the `active` flag
3. Post-processing effects captured `camera.value` before Vue's reactivity could update `activeCamera`
4. OrbitControls would modify the user's camera, but effects rendered with the old default camera

This resulted in broken interactivity and incorrect rendering in post-processing scenes.

#### 1. Async Renderer Initialization (`useRendererManager.ts`)

```typescript
// Initialize renderer asynchronously (required for WebGPU in Three.js r181+)
const initializeRenderer = async () => {
  if (isRenderer(renderer)) {
    await renderer.init()
    isInitialized.value = true
    readyEventHook.trigger(renderer)
  }
}
```

All renderer operations now wait for `isInitialized` before executing, preventing errors on uninitialized WebGPU contexts.

#### 2. Deferred Scene Mounting (`Context.vue`)

```typescript
renderer.onReady(() => {
  // Now that renderer is initialized, mount the actual scene with slots
  mountCustomRenderer(context.value, false)
  emit('ready', context.value)
})
```

The scene is only mounted after the renderer signals it's ready, ensuring all operations occur on an initialized renderer.

#### 3. Camera Registration Priority (`nodeOps.ts`)

```typescript
if (isTresCamera(child)) {
  // Register camera as active so it becomes the primary camera immediately
  // This ensures user cameras take precedence over the default camera
  context.camera?.registerCamera(child, true)
}
```

**This is the critical fix for post-processing.** When a user defines a camera in their template (e.g., `<TresPerspectiveCamera>`), it's now registered with `active=true`, immediately making it `cameras[0]` and therefore `activeCamera`. This ensures:

- Post-processing effects capture the correct camera from the start
- OrbitControls and effects operate on the same camera instance
- No race condition between camera registration and effect initialization

### Breaking Changes

None. This is a fix for compatibility with Three.js v0.181.0.

Closes TRES-183

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -267,6 +267,10 @@ For new features:
 - Consider if it belongs in core or a separate package
 - Be open to feedback and alternative approaches
 
+### Issue Triage
+
+To keep the issue tracker healthy and actionable, items marked with any of these labels — `needs reproduction`, `waiting for author`, or `needs-translations` — may be closed by the Tres team if there’s no activity for more than 30 days. If you later have new information (e.g., a minimal reproduction, updated context, or translations), please comment with the update and we’ll happily reopen.
+
 ## 💬 Community
 
 - Join our [Discord server](https://tresjs.org/discord) for questions and discussions


packages/core/src/composables/useRenderer/useRendererManager.ts
```diff
@@ -270,14 +270,55 @@ export function useRendererManager(
...
+  const errorEventHook = createEventHook<Error>()
   let hasTriggeredReady = false
 
-  if (isRenderer(renderer)) {
-    // Initialize the WebGPU context
-    renderer.init()
-    readyEventHook.trigger(renderer)
+  // Track whether the renderer has been initialized (important for WebGPU)
+  const isInitialized = ref(false)
+  const initializationError = ref<Error | null>(null)
```
Tinoooo: **suggestion:** I suggest not limiting this to initialisation errors to not block us in case we want to introduce other errors. Therefore I suggest naming this error and setting it's type to a renderer specific error type. That way, users can differentiate the errors when they are emitted from the TresCanvas component by using `instanceof`.

packages/core/src/composables/useRenderer/useRendererManager.ts
```diff
@@ -270,14 +270,55 @@ export function useRendererManager(
...
+        await renderer.init()
+        isInitialized.value = true
+        readyEventHook.trigger(renderer)
+      }
+      else {
+        // WebGLRenderer is ready immediately (no async init needed)
+        isInitialized.value = true
+        // Still need to trigger ready for WebGLRenderer for backward compatibility
+        readyEventHook.trigger(renderer)
+      }
```
Tinoooo: **suggestion(non-blocking):** The same things happen in the if block and in the else block. Can be unified.

packages/core/src/composables/useRenderer/useRendererManager.ts
```diff
@@ -270,14 +270,55 @@ export function useRendererManager(
...
+  const initializationError = ref<Error | null>(null)
+
+  // Initialize renderer asynchronously (required for WebGPU in Three.js r181+)
+  const initializeRenderer = async () => {
+    try {
+      if (isRenderer(renderer)) {
+        // WebGPU renderer requires awaiting init() before any operations
+        await renderer.init()
+        isInitialized.value = true
+        readyEventHook.trigger(renderer)
```
Tinoooo: **question(non-blocking):** Are we sure that the ready event can be triggered here safely? Iirc it was meant to be triggered when the renderer is ready and the canvas area is larger than 0. Can this be assumed here?

packages/core/src/core/nodeOps.ts
```diff
@@ -110,7 +110,9 @@ export const nodeOps = ({
     parent = unboxTresPrimitive(parentInstance)
 
     if (isTresCamera(child)) {
-      context.camera?.registerCamera(child)
+      // Register camera as active so it becomes the primary camera immediately
+      // This ensures user cameras take precedence over the default camera
+      context.camera?.registerCamera(child, true)
```
Tinoooo: **issue:** This change works around a shortcoming introduced in Context.vue. The default camera is always added even if the user added their own camera as the custom renderer is mounted after the check for an active camera. Therefore the component has no chance in detecting the camera that was added by the user. We should check after the mounting of the custom renderer. Otherwise there will always be an additional camera in the scene even if a user defined one is there.

packages/core/src/composables/useRenderer/useRendererManager.ts
alvarosabu: @Tinoooo good point, I added it here because it was already there in L278.

I see we have also another call in the watch for sizes. Do we need both?

Tinoooo: That's what I was wondering about. I'd say we don't need it. I removed it in [my PR](https://github.com/Tresjs/tres/pull/1278). But I'm also a little afraid of missing an edge case here. 😁 We really need more tests. 😁

Tinoooo (approved):

PR #1278, opened 2026-01-20
fix: 183 tres webgpu broken on three v181  tino

Tinoooo:
suggested changes concerning #1275


packages/core/src/components/Context.vue
```diff
@@ -232,6 +228,10 @@ renderer.onReady(() => {
   // Now that renderer is initialized, mount the actual scene with slots
   mountCustomRenderer(context.value, false)
   emit('ready', context.value)
+
+  if (!activeCamera.value) {
```
alvarosabu, line 232: Appreciation:

@Tinoooo oh God how did I missed this 🤦🏼

packages/core/src/composables/useRenderer/useRendererManager.ts
```diff
@@ -270,36 +271,34 @@ export function useRendererManager(
...
   const readyEventHook = createEventHook<TresRenderer>()
-  const errorEventHook = createEventHook<Error>()
+  const errorEventHook = createEventHook<TresRendererError>()
   let hasTriggeredReady = false
 
   // Track whether the renderer has been initialized (important for WebGPU)
-  const isInitialized = ref(false)
-  const initializationError = ref<Error | null>(null)
+  const isRendererInitialized = ref(false)
+  const error = ref<TresRendererError | null>(null)
```
alvarosabu: Question:

Why simplifying one (initializationError -> error) and making the other verbose (isInitialized -> isRendererInitialized) ?

Can we keep the first constant minimal but precise? It's repeating the surrounding context without any real benefit.

alvarosabu (changes requested):
Hey @Tinoooo thanks for reviewing and solving the camera activation more elegantly.

One naming note: Totally get your preference for explicit names. In the useRendererManager composable, the context already establishes ‘renderer,’ so adding it inside variables just makes them longer without improving clarity, same reason you probably renamed initializationError to error.

packages/core/src/composables/useRenderer/useRendererManager.ts
Tinoooo: My thinking was that `error` should be able to hold any error to be future-proof and `isRendererIntiialized` should represent what actually has been initialized.
Yes, I changed it back to the minimal variant. One could also read it as "the renderer manager is initialized" so that's fine. If we were very very consistent, this would even be `isInstanceIntialized`, but I guess this is too confusing. 😁

packages/core/src/index.ts
```diff
@@ -11,6 +11,8 @@ export type { TresCustomRendererOptions } from './core/nodeOps'
 export * from './devtools'
 export * from './directives'
 export * from './types'
+export * from './utils/error'
+export * from './utils/error'
```
Jungzl: This seems to be a duplicate export, right?

alvarosabu: @Tinoooo can you quickly update it, from my side is GTG

Tinoooo: Yes, I removed it.

PR #1280, opened 2026-01-22
chore(rapier): 205 move repo to monorepo

alvarosabu:
CLOSES TRES-205


alvarosabu:
> [BLOCKING] We don't need the vitepress docs I think is best to delete them.

@JaimeTorrealba I left them there in case you guys want to have them as reference for the docs on Nuxt, which we should do in a separate PR. Let me know if you still want them to be removed or not.

JaimeTorrealba:
I would prefer to be deleted, we got the other repo, I'll check that one @alvarosabu

alvarosabu:
> I would prefer to be deleted, we got the other repo, I'll check that one @alvarosabu

Done @JaimeTorrealba

Neosoulink:
I was having a bit of trouble launching the `package` and `playground`. Is it possible to create a doc for that?

> I know it's not related to this package/PR, but just a suggestion...

Neosoulink:
I'm noticing some issues with rapier examples, like in the following:

https://github.com/user-attachments/assets/1d3e4065-52ae-45b2-a32f-578f37b43af8

PR #1281, opened 2026-01-23
feat(core): fragment instance support

Tinoooo:


packages/core/src/core/nodeOps.test.ts
```diff
@@ -1571,6 +1571,191 @@ describe('nodeOps', () => {
       }
     })
   })
+
+  describe('createText (fragment anchors)', () => {
```
Tinoooo, line 1575: **fyi:** These tests are here to be uniform with the existing ones. I think eventually we should have tests that test from the user perspective. These test implementation details too much.

alvarosabu:
@Tinoooo i tried running the playground but I get a blank canvas? 
Screenshot 2026-01-26 at 16 27 45

Tinoooo:
> @Tinoooo i tried running the playground but I get a blank canvas? Screenshot 2026-01-26 at 16 27 45

Yes you are right. This happened after the merge from main. Seems like there have to be some adjustments for that. I try to work on that whenever I find time.

Tinoooo:
@alvarosabu I found the issue. The watcher that triggers the ready event was created too late, so it did not trigger in some cases. I fixed it in https://github.com/Tresjs/tres/pull/1281/changes/b58b785b38698cbb013213a8c21455efce645fd7. Sorry for not merging in main before requesting the review. I simply forgot.

packages/core/src/core/nodeOps.ts
```diff
@@ -93,18 +135,38 @@ export const nodeOps = ({
...
+      child.__tres.parent = parentInstance
+
+      if (parentInstance.__tres.objects) {
+        // Use findIndex with identity check to handle mixed TresTextNode/TresInstance arrays
+        const insertIndex = anchor ? parentInstance.__tres.objects.findIndex(obj => obj === anchor) : -1
+        if (insertIndex >= 0) {
+          parentInstance.__tres.objects.splice(insertIndex, 0, child as unknown as TresInstance)
+        }
+        else {
+          parentInstance.__tres.objects.push(child as unknown as TresInstance)
```
alvarosabu, line 171: @Tinoooo mmm is possible that the text/comment nodes hold parent references via `__tres.parent` and that is creating circular refs that block the GC.

`disposeObject3D` will dispose child objects, but I guess the arays of `__tres.objects` are never cleaned

Tinoooo: I don't think that's an issue. The elements in the array are just references. Also, the heap does not grow faster than it should (or does on main).

The problem was the handling of text nodes. Sometimes (I can't pinpoint in which cases), there is already a text node when a component is mounted. Vue wants to clear the fragment, but since the object array has no elements, the end text node cannot be found. The endless loop happens in [removeFragment](https://github.com/vuejs/core/blob/0b30b7d3be5185041b790286da8402dd10503dbb/packages/runtime-core/src/renderer.ts#L2295). I adjusted the code now, so that if Context.vue is in the process of unmounting, no new text nodes are created. It seems to fix the issue. Very hard to debug 😟.

alvarosabu: I see, buff nice catch man, hard one

Issue #1283, opened 2026-01-23
Support for fragment instances


Tinoooo:
### Description

As a developer using Tresjs, I want to use multi-root node components, i.e., fragments, because I am used to that from Vue.js's default renderer and it is frustrating to not have this feature.

The lack of the support (and the frustration 😁) becomes obvious in multiple bug reports.
#1245 
#746 
#1005 

### Suggested solution

Make our custom renderer support fragment instances.

### Validations

- [x] I agree to follow this project's [Code of Conduct](https://github.com/Tresjs/tres/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/Tresjs/.github/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://tresjs.org/guide).
- [x] Check that there isn't [already an issue](https://github.com/tresjs/tres/issues) that reports the same bug to avoid creating a duplicate.

PR #1321, opened 2026-02-03
chore(core): user perspective tests

Tinoooo:
This PR is a first start to having tests that mimic user behavior. I think we should prefer this style to avoid testing implementation details. The PR is also here to show how we could mock WebGL. My plan is to add more tests in the future to gradually replace the tests that test implementation details (where possible). The PR is also a little bit of a follow-up to #1283.


packages/core/src/components/TresScene.test/hierarchy.test.ts
```diff
@@ -0,0 +1,88 @@
...
+  let sceneWrapper: { unmount: () => void }
+  let group: THREE.Group
+
+  beforeAll(async () => {
+    const { createScene } = await initializeSceneCreator()
+    const result = await createScene(() => h(Component))
+    sceneWrapper = result.sceneWrapper
+    scene = result.context.scene.value
+    group = scene.children[0] as THREE.Group
+  })
```
alvarosabu, line 57: Question:

Are the tests properly isolated with `beforeAll` ? As far as I remember this will mantain a shared state across tests, so if one tests modifies it subsequent tests might be affected. Maybe we should use `beforeEach`?

Tinoooo: Yes, it does remain the shared state. I used `beforeAll` to have a tiny performance gain as the tests only check and don't modify the scene. But we can, for the sake of best practices, also use `beforeEach` here. What would you prefer?

alvarosabu: @Tinoooo its ok then, we just need to be aware in the future if that changes.

PR #1325, opened 2026-02-05
feat(cientos): add useBVH composable for raycasting optimization

alvarosabu:
## Summary

Introduces `useBVH` composable for @tresjs/cientos that dramatically improves raycasting performance by computing Bounding Volume Hierarchy (BVH) structures on mesh geometries.

### Features
- **useBVH composable** - Apply BVH optimization to any Object3D or scene
- **applyBVHWhenReady helper** - Async helper for applying BVH to loaded models
- **Reactive enabled option** - Toggle BVH on/off reactively
- **firstHitOnly mode** - Faster single-hit raycasting with proper raycaster settings (near/far/layers)
- **Debug visualization** - Optional BVH bounding box visualization
- **Configurable options** - splitStrategy, maxDepth, maxLeafSize, indirect
- **Proper memory management** - MeshBVH disposal on unmount/disable to prevent memory leaks

### Changes
- New `useBVH` composable in `packages/cientos/src/core/performance/`
- Added `three-mesh-bvh` dependency
- Playground demo with BVH performance example
- Full documentation in cientos-docs

## Test plan
- [x] BVH applies correctly to meshes and groups
- [x] firstHitOnly mode respects raycaster near/far/layers
- [x] Memory properly cleaned up on unmount
- [x] Debug helpers display correctly
- [x] Reactive enabled toggle works

Closes #61

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/apps/cientos-docs/app/components/debug-performance/UseBVH.vue
+++ b/apps/cientos-docs/app/components/debug-performance/UseBVH.vue
@@ -1,6 +1,6 @@
 <script setup lang="ts">
-import { TresCanvas, type TresObject3D } from '@tresjs/core'
-import { OrbitControls, useGLTF, useBVH } from '@tresjs/cientos'
+import { TresCanvas } from '@tresjs/core'
+import { OrbitControls, useBVH, useGLTF } from '@tresjs/cientos'
 import { useControls } from '@tresjs/leches'
 
 const uuid = inject('uuid')
@@ -15,15 +15,14 @@ const { enabled, debug } = useControls({
 }, { uuid })
 
 // BVH composable with reactive controls
-const { applyBVHWhenReady } = useBVH({
-  enabled,
-  debug,
-  verbose: false,
-})
-
-// Apply BVH when model loads
-const modelScene = computed(() => model.value?.scene)
-applyBVHWhenReady(modelScene as Ref<TresObject3D | null | undefined>)
+useBVH(
+  () => model.value?.scene,
+  {
+    enabled,
+    debug,
+    verbose: false,
+  },
+)
 </script>
 
 <template>
--- a/apps/cientos-docs/content/2.api/6.debug-performance/use-bvh.md
+++ b/apps/cientos-docs/content/2.api/6.debug-performance/use-bvh.md
@@ -30,11 +30,7 @@ BVH (Bounding Volume Hierarchy) is a spatial data structure that organizes geome
 import { useGLTF, useBVH } from '@tresjs/cientos'
 
 const { state: model } = useGLTF('/models/complex-model.glb')
-const { applyBVHWhenReady } = useBVH()
-
-// Apply BVH when model loads
-const modelScene = computed(() => model.value?.scene)
-applyBVHWhenReady(modelScene)
+useBVH(() => model.value?.scene)
 </script>
 
 <template>
@@ -50,14 +46,14 @@ Enable debug mode to visualize the BVH bounding boxes:
 <script setup lang="ts">
 import { useGLTF, useBVH } from '@tresjs/cientos'
 
-const { applyBVHWhenReady } = useBVH({
-  debug: true, // Show BVH bounding boxes
-})
-
 const { state: model } = useGLTF('/models/model.glb')
 
-const modelScene = computed(() => model.value?.scene)
-applyBVHWhenReady(modelScene)
+useBVH(
+  () => model.value?.scene,
+  {
+    debug: true, // Show BVH bounding boxes
+  }
+)
 </script>
 ```
 
@@ -72,14 +68,15 @@ import { useGLTF, useBVH } from '@tresjs/cientos'
 
 const bvhEnabled = ref(true)
 
-const { applyBVHWhenReady } = useBVH({
-  enabled: bvhEnabled,
-})
-
 const { state: model } = useGLTF('/models/model.glb')
 
-const modelScene = computed(() => model.value?.scene)
-applyBVHWhenReady(modelScene)
+useBVH(
+  () => model.value?.scene,
+  {
+    enabled: bvhEnabled,
+  }
+)
+
 </script>
 
 <template>
@@ -90,23 +87,6 @@ applyBVHWhenReady(modelScene)
     <primitive v-if="model" :object="model.scene" />
   </div>
 </template>
-```
-
-### Manual Control
-
-Apply BVH directly to objects:
-
-```vue{8}
-<script setup lang="ts">
-import { useGLTF, useBVH } from '@tresjs/cientos'
-
-const { applyBVH } = useBVH()
-const { state: model } = useGLTF('/models/model.glb')
-
-watch(() => model.value, (newModel) => {
-  if (newModel) applyBVH(newModel.scene)
-})
-</script>
 ```
 
 ## Options
@@ -140,15 +120,6 @@ These options configure how the BVH is built. Changing them after creation has n
 - **AVERAGE** - Balanced build time and runtime performance.
 - **CENTER** - Fastest to build, slower at runtime.
 
-## Return Values
-
-| Name | Type | Description |
-| :--- | :--- | :--- |
-| **processedMeshes** | `DeepReadonly<Ref<ProcessedMesh[]>>` | Readonly array of meshes with BVH applied. |
-| **applyBVH** | `(object: Object3D) => void` | Apply BVH to an object and all its mesh children. |
-| **applyBVHWhenReady** | `(objectRef: Ref<Object3D \| null>) => void` | Watch a ref/computed and apply BVH when the object becomes available. |
-| **removeBVH** | `() => void` | Remove BVH from all processed meshes. |
-
 ## Advanced Usage
 
 ### Fine-tuning Performance
@@ -156,12 +127,15 @@ These options configure how the BVH is built. Changing them after creation has n
 Adjust BVH construction parameters for your use case:
 
 ```ts
-const { applyBVHWhenReady } = useBVH({
-  splitStrategy: 'SAH',     // Best runtime performance
-  maxDepth: 30,             // Shallower tree (faster build)
-  maxLeafSize: 5,           // Smaller leaves (better culling)
-  verbose: true,            // Debug construction
-})
+useBVH(
+  target,
+  {
+    splitStrategy: 'SAH', // Best runtime performance
+    maxDepth: 30, // Shallower tree (faster build)
+    maxLeafSize: 5, // Smaller leaves (better culling)
+    verbose: true, // Debug construction
+  }
+)
 ```
 
 ### First Hit Only Mode
@@ -169,9 +143,12 @@ const { applyBVHWhenReady } = useBVH({
 When you only need the closest intersection (e.g., mouse picking):
 
 ```ts
-const { applyBVHWhenReady } = useBVH({
-  firstHitOnly: true,  // Uses raycastFirst internally
-})
+useBVH(
+  target,
+  {
+    firstHitOnly: true, // Uses raycastFirst internally
+  }
+)
 ```
 
 This is significantly faster than computing all intersections when you only need one.
@@ -190,7 +167,6 @@ This is significantly faster than computing all intersections when you only need
 2. **SAH strategy** gives best runtime performance for production
 3. **Adjust maxLeafSize** based on triangle density (smaller for dense meshes)
 4. **Enable debug mode** during development to verify BVH coverage
-5. **Apply to loaded models only** - use `applyBVHWhenReady` with async loaders
 
 ## Integration with useGLTF
 
@@ -198,23 +174,16 @@ Perfect pairing with `useGLTF` for optimized model loading:
 
 ```vue
 <script setup lang="ts">
-import { useGLTF, useBVH } from '@tresjs/cientos'
+import { useBVH, useGLTF } from '@tresjs/cientos'
 
 const { state: model } = useGLTF('/models/high-poly-model.glb', {
   draco: true
 })
 
-const { applyBVHWhenReady, processedMeshes } = useBVH({
-  splitStrategy: 'SAH',
-})
-
-const modelScene = computed(() => model.value?.scene)
-applyBVHWhenReady(modelScene)
-
-// processedMeshes updates when BVH is applied
-watch(processedMeshes, (meshes) => {
-  console.log(`BVH applied to ${meshes.length} meshes`)
-})
+useBVH(
+  () => model.value?.scene,
+  { splitStrategy: 'SAH' }
+)
 </script>
 ```
 
--- a/apps/playground/src/pages/cientos/performance/bvh/ModelWithBVH.vue
+++ b/apps/playground/src/pages/cientos/performance/bvh/ModelWithBVH.vue
@@ -1,6 +1,5 @@
 <script setup lang="ts">
-import { useGLTF, useBVH, Html } from '@tresjs/cientos'
-import { TresObject3D } from '@tresjs/core'
+import { Html, useBVH, useGLTF } from '@tresjs/cientos'
 import { useControls } from '@tresjs/leches'
 
 const { state: model } = useGLTF('/models/Artificer.glb', {
@@ -14,20 +13,21 @@ const { enabled, debug } = useControls({
   debug: false,
 }, { uuid })
 
-const { applyBVHWhenReady } = useBVH({
-  enabled,
-  debug,
-})
-
-const modelScene = computed(() => model.value?.scene)
-applyBVHWhenReady(modelScene as Ref<TresObject3D | null | undefined>)
-
+useBVH(
+  () => model.value?.scene,
+  {
+    enabled,
+    debug,
+  },
+)
 const isHovering = ref(false)
 
 const handleClick = (event: PointerEvent) => {
+  // eslint-disable-next-line no-console
   console.log('clicked', event)
 }
 </script>
+
 <template>
   <primitive
     v-if="model?.scene"
--- a/packages/cientos/src/core/performance/useBVH.ts
+++ b/packages/cientos/src/core/performance/useBVH.ts
@@ -1,9 +1,9 @@
 /* eslint-disable no-console */
-import { type DeepReadonly, type MaybeRefOrGetter, onUnmounted, readonly, type Ref, ref, toValue, unref, watch } from 'vue'
+import { type MaybeRefOrGetter, onUnmounted, toValue, watch } from 'vue'
 import { whenever } from '@vueuse/core'
 import { acceleratedRaycast, BVHHelper, MeshBVH } from 'three-mesh-bvh'
 import { type BufferGeometry, Group, Mesh, type Object3D, SkinnedMesh } from 'three'
-import type { TresObject3D } from '@tresjs/core'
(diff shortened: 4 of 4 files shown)


apps/cientos-docs/content/2.api/6.debug-performance/use-bvh.md
```diff
@@ -0,0 +1,238 @@
+---
+title: useBVH
+description: Exponentially speed up raycasting with Bounding Volume Hierarchy (BVH) optimization.
```
Tinoooo: **suggestion:** I suggest omitting the word exponential here. An improvement of the complexity from O(n) to O(log n) is sub-exponential.

```suggestion
description: Speed up raycasting with Bounding Volume Hierarchy (BVH) optimization.
```

apps/cientos-docs/content/2.api/6.debug-performance/use-bvh.md
```diff
@@ -0,0 +1,238 @@
...
+---
+
+::SceneControlsWrapper
+  ::DebugPerformanceUseBvh
+  ::
+::
+
+A composable that dramatically improves raycasting performance by building a Bounding Volume Hierarchy (BVH) for your meshes. This can speed up raycasting by **orders of magnitude**, especially for complex models with many triangles.
+
+Built on top of [three-mesh-bvh](https://github.com/gkjohnson/three-mesh-bvh) by Garrett Johnson.
```
Tinoooo: **suggestion(non-blocking):**

```suggestion
Built on top of [three-mesh-bvh](https://github.com/gkjohnson/three-mesh-bvh) by [Garrett Johnson](https://github.com/gkjohnson).
```

packages/cientos/src/core/performance/useBVH.ts
```diff
@@ -0,0 +1,380 @@
...
+
+  // Sync isEnabled with reactive enabled option
+  watchEffect(() => {
+    isEnabled.value = toValue(enabled)
+  })
+
+  // Sync isDebug with reactive debug option
+  watchEffect(() => {
+    isDebug.value = toValue(debug)
+  })
```
Tinoooo: **suggestion** I suggest using the passed options directly without syncing them manually. The watchers can be avoided.

packages/cientos/src/core/performance/useBVH.ts
```diff
@@ -0,0 +1,380 @@
...
+    processedMeshes.value.forEach(removeBVHFromMesh)
+    processedMeshes.value = []
+  }
+
+  /**
+   * Toggle BVH optimization
+   */
+  const setBVHEnabled = (enabled: boolean): void => {
+    isEnabled.value = enabled
+  }
```
Tinoooo: **suggestion:** I suggest not providing such a method. If users want to toggle this, they should provide a ref in the options. We have Vue's reactivity features, so we should use them.

packages/cientos/src/core/performance/useBVH.ts
```diff
@@ -0,0 +1,380 @@
...
+
+  /**
+   * Helper for async model loading - watches a getter and applies BVH when truthy
+   * @param objectGetter - A getter function that returns the Object3D when ready
+   */
+  const applyBVHWhenReady = (objectGetter: () => TresObject3D | Object3D | null | undefined): void => {
+    watchEffect(() => {
```
Tinoooo: **suggestion:** Watchers should only be created within setup. Calling this in an exposed method might mislead users into calling this outside of the setup scope. Would it make sense to have the watcher in the root of the composable and pass to the `objectGetter`(I'd prefer a ref or computed) instead?

packages/cientos/src/core/performance/useBVH.ts
```diff
@@ -0,0 +1,380 @@
...
+    meshes.forEach((mesh, index) => {
+      // Skip if already processed
+      const alreadyProcessed = processedMeshes.value.find(pm => pm.mesh === mesh)
+      if (alreadyProcessed) {
+        if (verbose) {
+          console.log(`BVH: Mesh ${index + 1} already processed, skipping`)
+        }
+        return
+      }
+
```
Tinoooo: **suggestion(non-blocking):** If the console message is not absolutely necessary I suggest looping through the processedMeshes only once instead of for each mesh.

```suggestion
    const unprocessedMeshes = meshes.filter(mesh => !processedMeshes.value.some(pm => pm.mesh === mesh))
    unprocessedMeshes.forEach((mesh, index) => {
```

packages/cientos/src/core/performance/useBVH.ts
```diff
@@ -0,0 +1,380 @@
...
+  watch(isEnabled, (newEnabled) => {
+    if (!newEnabled) {
+      removeBVH()
+    }
+  })
+
+  // Watch for debug state changes
+  watch(isDebug, () => {
+    updateDebugHelpers()
+  })
```
Tinoooo: **suggestion:** I'm not entirely sure, but I guess this should be triggered immediately.
```suggestion
  }, { immediate: true })
```

packages/cientos/src/core/performance/useBVH.ts
```diff
@@ -0,0 +1,380 @@
...
+    enabled = true,
+    firstHitOnly = false,
+    splitStrategy = 'SAH',
+    verbose = false,
+    setBoundingBox = true,
+    maxDepth = 40,
+    maxLeafSize = 10,
+    indirect = false,
+    debug = false,
+  } = options
```
Tinoooo: **nitpick:** 
```suggestion
export const useBVH = ({
  enabled = true,
  firstHitOnly = false,
  splitStrategy = 'SAH',
  verbose = false,
  setBoundingBox = true,
  maxDepth = 40,
  maxLeafSize = 10,
  indirect = false,
  debug = false,
}: UseBVHOptions = {}) => {
```

packages/cientos/src/core/performance/useBVH.ts
```diff
@@ -0,0 +1,380 @@
...
+  // Cleanup on unmount
+  onUnmounted(() => {
+    removeBVH()
+  })
+
+  return {
+    // State
+    isEnabled,
+    isDebug,
+    processedMeshes: processedMeshes as Ref<readonly ProcessedMesh[]>,
```
Tinoooo: **suggestion:** I suggest making this readonly not only type-wise.

```suggestion
    processedMeshes: readonly(processedMeshes),
```

packages/cientos/src/core/performance/useBVH.ts
```diff
@@ -0,0 +1,380 @@
...
+    // Methods
+    applyBVH,
+    applyBVHWhenReady,
+    removeBVH,
+    setBVHEnabled,
+
+    // Computed info
+    get meshCount() {
+      return processedMeshes.value.length
+    },
```
Tinoooo: **suggestion:** I suggest  using a computed instead to profit from vue's dependency tracking optimizations. I'm not entirely sure whether the average user of the composable needs this value. If not, I event suggest omitting this value entirely.

packages/cientos/src/core/performance/useBVH.ts
```diff
@@ -0,0 +1,380 @@
...
+      }
+    })
+  }
+
+  // Watch for enabled state changes
+  watch(isEnabled, (newEnabled) => {
+    if (!newEnabled) {
+      removeBVH()
+    }
+  })
```
Tinoooo: **nitpick:**
```suggestion
  whenever(() => !unref(enabled), () => {
    removeBVH()
  })
```

packages/cientos/src/core/performance/useBVH.ts
```diff
@@ -0,0 +1,380 @@
...
+    })
+
+    return meshes
+  }
+
+  /**
+   * Apply BVH optimization to a mesh (Mesh or SkinnedMesh)
+   */
+  const applyBVHToMesh = (mesh: Mesh | SkinnedMesh): ProcessedMesh | null => {
+    if (!mesh.geometry || !mesh.geometry.attributes.position) {
```
Tinoooo: **nitpick:**
```suggestion
    if (!mesh.geometry?.attributes.position) {
```

packages/cientos/src/core/performance/useBVH.ts
```diff
@@ -0,0 +1,380 @@
...
+          const boundsTree = this.geometry.boundsTree
+          if (boundsTree) {
+            // Check layer intersection
+            if (!this.layers.test(raycaster.layers)) {
+              return
+            }
+
+            const material = Array.isArray(this.material) ? this.material[0] : this.material
+            const side = material?.side ?? raycaster.params.Mesh?.side
+            const hit = (boundsTree as MeshBVH).raycastFirst(
```
Tinoooo: **suggestion:** I suggest checking the type instead of casting it.

```suggestion
            if (!(boundsTree instanceof MeshBVH)) {
              return
            }
            const hit = boundsTree.raycastFirst(
```

packages/cientos/src/core/performance/useBVH.ts
```diff
@@ -0,0 +1,380 @@
...
+  /**
+   * Create debug helper for a processed mesh
+   */
+  const createDebugHelper = (processedMesh: ProcessedMesh): void => {
+    if (processedMesh.debugHelper) {
+      return // Already has helper
+    }
+
+    // Use the boundsTree directly for the helper
+    const helper = new BVHHelper(processedMesh.mesh as unknown as Object3D, processedMesh.boundsTree, maxDepth)
```
Tinoooo: **suggestion:** The casting should not be required. If it is really required I suggest using our type guards instead.
```suggestion
    const helper = new BVHHelper(processedMesh.mesh, processedMesh.boundsTree, maxDepth)
```

packages/cientos/src/core/performance/useBVH.ts
```diff
@@ -0,0 +1,380 @@
...
+  indirect?: boolean
+}
+
+export interface UseBVHOptions extends BVHOptions {
+  /** Enable/disable BVH optimization, default: true. Reactive - can be a ref or getter. */
+  enabled?: MaybeRefOrGetter<boolean>
+  /** Use .raycastFirst to retrieve hits which is generally faster, default: false. Static. */
+  firstHitOnly?: boolean
+  /** Show debug visualization of BVH bounding boxes, default: false. Reactive - can be a ref or getter. */
+  debug?: MaybeRefOrGetter<boolean>
```
Tinoooo: **nitpick:** I suggest not meintioning the reacitivity in the comments as it is already obvious in the types.

```suggestion
  /** Split strategy, default: SAH (slowest to construct, fastest runtime, least memory). */
  splitStrategy?: 'CENTER' | 'AVERAGE' | 'SAH'
  /** Print out warnings encountered during tree construction, default: false. */
  verbose?: boolean
  /** If true then the bounding box for the geometry is set once the BVH has been constructed, default: true. */
  setBoundingBox?: boolean
  /** The maximum depth to allow the tree to build to, default: 40. */
  maxDepth?: number
  /** The number of triangles to aim for in a leaf node, default: 10. */
  maxLeafSize?: number
  /** If false then an index buffer is created if it does not exist and is rearranged. */
  indirect?: boolean
}

export interface UseBVHOptions extends BVHOptions {
  /** Enable/disable BVH optimization, default: true. */
  enabled?: MaybeRefOrGetter<boolean>
  /** Use .raycastFirst to retrieve hits which is generally faster, default: false. */
  firstHitOnly?: boolean
  /** Show debug visualization of BVH bounding boxes, default: false. */
  debug?: MaybeRefOrGetter<boolean>
```

packages/cientos/src/core/performance/useBVH.ts
```diff
@@ -0,0 +1,380 @@
...
+  /**
+   * Update debug visualization for all processed meshes
+   */
+  const updateDebugHelpers = (): void => {
+    if (isDebug.value) {
+      processedMeshes.value.forEach(createDebugHelper)
+    }
+    else {
+      processedMeshes.value.forEach(removeDebugHelper)
+    }
```
Tinoooo:
```suggestion
      processedMeshes.value.forEach(isDebug.value ? createDebugHelper : removeDebugHelper)
```

apps/playground/src/pages/cientos/performance/bvh/ModelWithBVH.vue
```diff
@@ -0,0 +1,50 @@
...
+  debug,
+})
+
+applyBVHWhenReady(() => model.value?.scene)
+
+const isHovering = ref(false)
+
+const handleClick = (event: PointerEvent) => {
+  console.log('clicked', event)
+}
```
Tinoooo, line 28: **question:** The hover and click events trigger also when clicking outside (probably inside the bounding box) of the samurai figure.  Is that wanted behavior? If so, a comment in the demo code would be nice so nobody thinks the raycasting is broken.

packages/cientos/src/core/performance/useBVH.ts
Tinoooo, line 1: **suggestion(non-blocking):** Tests would be nice.

apps/cientos-docs/app/components/debug-performance/UseBVH.vue
Tinoooo, line 1: **issue:** The file has linting and typescript errors

apps/cientos-docs/content/2.api/6.debug-performance/use-bvh.md
```diff
@@ -0,0 +1,238 @@
...
+### Basic Usage
+
+```vue{2,5-8,10}
+<script setup lang="ts">
+import { useGLTF, useBVH } from '@tresjs/cientos'
+
+const { state: model } = useGLTF('/models/complex-model.glb')
+const { applyBVHWhenReady } = useBVH({
+  enabled: true,
+})
```
Tinoooo: **nitpick:** default value can be omitted.
```suggestion
const { applyBVHWhenReady } = useBVH()
```

apps/cientos-docs/content/2.api/6.debug-performance/use-bvh.md
```diff
@@ -0,0 +1,238 @@
...
+### With Debug Visualization
+
+Enable debug mode to visualize the BVH bounding boxes:
+
+```vue{4-7}
+<script setup lang="ts">
+import { useGLTF, useBVH } from '@tresjs/cientos'
+
+const { applyBVHWhenReady } = useBVH({
+  enabled: true,
```
Tinoooo: **nitpick:** default value can be omitted.
```suggestion
```

apps/cientos-docs/content/2.api/6.debug-performance/use-bvh.md
```diff
@@ -0,0 +1,238 @@
...
+| **debug** | `MaybeRefOrGetter<boolean>` | `false` | Show debug visualization of BVH bounding boxes. |
+
+### Static Options (BVH Construction)
+
+These options configure how the BVH is built. Changing them after creation has no effect - toggle `enabled` off/on to rebuild with new values.
+
+| Option | Type | Default | Description |
+| :--- | :--- | :--- | :--- |
+| **firstHitOnly** | `boolean` | `false` | Use `raycastFirst` for better performance when only the first hit is needed. |
+| **splitStrategy** | `'CENTER' \| 'AVERAGE' \| 'SAH'` | `'SAH'` | BVH build strategy. SAH is slowest to construct but fastest at runtime. |
```
Tinoooo:
```suggestion
| **splitStrategy** | `'CENTER' \| 'AVERAGE' \| 'SAH'` | `'SAH'` | BVH build strategy. See section below.
```

packages/cientos/src/core/performance/useBVH.ts
alvarosabu: @Tinoooo  So I tried your suggestion to convert it into an arrow function + the readonly but that caused the famous `TS7056: The inferred type of this node exceeds the maximum length the compiler will serialize` so I kept the export function but I still desturctured it on the explicit interface. 

I guess the combination of `readonly()` + Threejs deep types make making explicit returns necessaries 😅. If you want to give it a try all yours but I would not block it for a nitpick tbh

apps/cientos-docs/app/components/debug-performance/UseBVH.vue
alvarosabu: The error is because the modelScene is not typed, I can only do a dirty assertion there

apps/playground/src/pages/cientos/performance/bvh/ModelWithBVH.vue
alvarosabu: I was able to reproduce it @Tinoooo, the accelerated raycast will also work on the triangles so its not necessarely the bounding box (if you check debug true you will see the boundboxes created)

I tried with another model and it doesnt happen, I believe the Samurai one might have some hidden geometries

packages/cientos/src/core/performance/useBVH.ts
Tinoooo: I just noticed we have to use toValue here, because enabled can be a getter function. I'll change it in a separate PR.

packages/cientos/src/core/performance/useBVH.ts
Tinoooo: There is still a watcher. I'll show you how I think it could work in a separate PR.

packages/cientos/src/core/performance/useBVH.ts
Tinoooo: The explicit return type is totally fine here. I'd even prefer it over an inferred one.

Yes, I'll give it a try. fyi: [nitpicks are non-blocking by default](https://conventionalcomments.org/#labels).

packages/cientos/src/core/performance/useBVH.ts
Tinoooo: Today typescript complains about the missing type on my machine too. Not sure why it is so inconsistent. I'll add a typeguard.

apps/cientos-docs/app/components/debug-performance/UseBVH.vue
Tinoooo: the linter gave me errors because of [perfectionist/sort-named-imports](https://perfectionist.dev/rules/sort-named-imports). fixed in seperate PR

apps/playground/src/pages/cientos/performance/bvh/ModelWithBVH.vue
Tinoooo: The new model is nice :) 👍

packages/cientos/src/core/performance/useBVH.ts
```diff
@@ -0,0 +1,297 @@
...
+  maxDepth?: number
+  /** The number of triangles to aim for in a leaf node, default: 10. */
+  maxLeafSize?: number
+  /** If false then an index buffer is created if it does not exist and is rearranged. */
+  indirect?: boolean
+}
+
+export interface UseBVHOptions extends BVHOptions {
+  /** Enable/disable BVH optimization, default: true. */
+  enabled?: MaybeRefOrGetter<boolean>
```
Tinoooo, line 28: **issue:** enabled is not really reactive. when going from false to true, the bvh is not built. One could argue, that the user needs to call the apply methods again, but it might be a bit unintuitive for the user. I addressed this in the other PR.

PR #1330, opened 2026-02-08
fix(rapier): colliders appear again on scene

JaimeTorrealba:
Beside fixing the colliders, I improve RigidBody and Gravity demos on playground with better props tweaking.

[UPDATE]:

I realize the problem with the reactivity of the colliders, in our RigidBody coore component we got a:
`<slot v-once></slot>` who was blocking the reactivity. Now I remember Nathan add this one to our codebase for one reason but no idea why. for noe the reactivity is back.

The reactivity of the Automatic collider is still missing but I think I can fix it very fast I know what the problem is, I'm going to address it in another PR

Closes TRES-245

### DIFF
--- a/apps/docs/content/3.api/1.components/tres-context.md
+++ b/apps/docs/content/3.api/1.components/tres-context.md
@@ -1,0 +1,42 @@
+---
+title: <TresCanvasContext />
+description: Internal context component used by TresCanvas for advanced setups.
+---
+
+## Component Overview
+
+`<TresCanvasContext />` is the internal component that powers `<TresCanvas />`. It mounts the TresJS renderer and scene using a **provided canvas element** instead of creating one for you.
+
+::note
+This component is exported for advanced use cases only. In most apps you should keep using `<TresCanvas />`.
+::
+
+## When to Use It
+
+Use `<TresCanvasContext />` only if you already own the WebGL canvas (or must integrate with a host framework that provides one) and you still want TresJS to manage the Three.js scene, render loop, and events.
+
+## Usage
+
+```vue [app.vue]
+<script setup lang="ts">
+import { ref } from 'vue'
+import { TresCanvasContext } from '@tresjs/core'
+
+const canvasRef = ref<HTMLCanvasElement>()
+</script>
+
+<template>
+  <canvas ref="canvasRef" class="w-full h-full">
+    <TresCanvasContext v-if="canvasRef" :canvas="canvasRef">
+      <TresPerspectiveCamera :position="[3, 3, 3]" />
+      <!-- Your scene content here -->
+    </TresCanvasContext>
+  </canvas>
+</template>
+```
+
+## Notes
+
+- You must pass a valid `canvas` element.
+- Canvas sizing and styling are **your** responsibility.
+- Props and events match `<TresCanvas />`, so you can reuse the same API surface.
--- a/packages/core/src/components/index.ts
+++ b/packages/core/src/components/index.ts
@@ -1,4 +1,5 @@
 import TresCanvas from './TresCanvas.vue'
+import Context from './Context.vue'
 
-export { TresCanvas }
+export { Context as TresCanvasContext, TresCanvas }
 export type { TresCanvasEmits, TresCanvasInstance, TresCanvasProps } from './TresCanvas.vue'
--- a/packages/core/src/utils/template-compiler-options.ts
+++ b/packages/core/src/utils/template-compiler-options.ts
@@ -1,5 +1,6 @@
 const whitelist = [
   'TresCanvas',
+  'TresCanvasContext',
   'TresLeches',
   'TresScene',
 ]


packages/rapier/src/components/RigidBody.vue
```diff
@@ -88,7 +89,10 @@ watch(bodyGroup, async (group) => {
     const collidersProps: ColliderProps[] = []
 
     for (const child of group.children) {
-      const _props = createColliderPropsFromObject(child, props.collider)
+      // Skip children without valid geometry (e.g., collider wrappers, empty Object3Ds)
+      if (!hasValidColliderGeometry(child as Object3D)) { continue }
```
alvarosabu, line 93: Appreciation:

Nice prevention to avoid creating colliders from empty wrappers and non-rendable objects.

packages/rapier/src/utils/collider.ts
```diff
@@ -1,6 +1,29 @@
...
+ * Returns true if the object has geometry with position attributes,
+ * or if it has descendants with geometry (non-empty bounding box).
+ *
+ * @param object {@link Object3D}
+ */
+export const hasValidColliderGeometry = (object?: Object3D): boolean => {
+  if (!object) { return false }
+
+  // Check if the object itself has geometry with positions
+  const geo = (object as unknown as { geometry?: { attributes?: { position?: { array?: ArrayLike<number> } } } }).geometry
```
alvarosabu: blocker:

Avoid type assertions chains, it's better to use a type guard like we do in the core

```ts
  function hasGeometry(object: Object3D): object is Object3D & { geometry:
  BufferGeometry } {                                                                
    return 'geometry' in object && (object as Mesh).geometry instanceof
  BufferGeometry
  }

  Then usage becomes simply:

  if (hasGeometry(child)) {
    // child.geometry is properly typed here
  }

```

packages/rapier/src/utils/collider.ts
```diff
@@ -1,6 +1,29 @@
...
+  if (geo?.attributes?.position?.array?.length) {
+    return true
+  }
+
+  // Check if the object or its descendants have valid geometry via bounding box
+  const boundingBox = new Box3().setFromObject(object)
+  // Box3.isEmpty() returns true if no points have been added to the box
+  return !boundingBox.isEmpty()
```
alvarosabu: Blocker:

`hasValidColliderGeometry` here creates a `Box3().setFromObject()` per
   child on every mount. 

Also`setFromObject` traverses the entire subtree of the object recursively which is quite resource expensive.

For a cheaper check you could just test if the object is a Mesh with a
  `non-empty `geometry:

```ts
  function hasValidColliderGeometry(object: Object3D): boolean {
    if ((object as Mesh).geometry?.attributes?.position) return true
    // check children only if needed
    return object.children.some(child => hasValidColliderGeometry(child))
  }
```

apps/playground/src/pages/rapier/CustomColliders.vue
```diff
@@ -0,0 +1,79 @@
...
+          <BallCollider
+            :args="[1, 1, 1]"
+            :friction="friction"
+            :mass="mass"
+            :restitution="restitution"
+            :density="density"
+          />
+        </RigidBody>
+
+        <RigidBody type="fixed" :restitution="restitution" @click="jump">
```
alvarosabu:
```suggestion
        <RigidBody type="fixed" restitution @click="jump">
```

apps/playground/src/pages/rapier/CustomColliders.vue
```diff
@@ -0,0 +1,79 @@
...
+</script>
+
+<template>
+  <TresLeches />
+  <TresCanvas v-bind="gl">
+    <TresPerspectiveCamera :position="[11, 25, 25]" :look-at="[0, 0, 0]" />
+    <OrbitControls />
+
+    <Suspense>
+      <Physics :debug="debug">
```
alvarosabu:
```suggestion
      <Physics debug>
```

apps/playground/src/pages/rapier/GravityDemo.vue
```diff
@@ -13,45 +13,62 @@ const gl = {
...
+    <TresPerspectiveCamera :position="[0, 0, 45]" :look-at="[0, 0, 0]" />
     <OrbitControls />
-    
+
     <Suspense>
-      <Physics debug :gravity="[0, gravityY, 0]">
-        <RigidBody>
-          <TresMesh :position="[0, 8, 0]">
-            <TresTorusGeometry />
+      <Physics :debug="debug" :gravity="[gravityX, gravityY, gravityZ]">
```
alvarosabu:
```suggestion
      <Physics debug :gravity="[gravityX, gravityY, gravityZ]">
```

apps/playground/src/pages/rapier/RigidBody.vue
```diff
@@ -11,22 +13,135 @@ const gl = {
...
+  <TresLeches />
   <TresCanvas v-bind="gl">
     <TresPerspectiveCamera :position="[11, 11, 11]" :look-at="[0, 0, 0]" />
     <OrbitControls />
 
     <Suspense>
-      <Physics debug>
-        <RigidBody>
-          <TresMesh :position="[0, 8, 0]">
+      <Physics :debug="debug">
```
alvarosabu:
```suggestion
      <Physics debug>
```

JaimeTorrealba:
I think I address all your concerns @alvarosabu 
Small one this doesn't work `<Physics debug>` it need to be bound `<Physics :debug>`

Issue #1332, opened 2026-02-09
PointerLockControls emits too many events


Tinoooo:
### Describe the bug

The component emits events too often, because [it registers event listeners on each click](https://github.com/Tresjs/tres/blob/5882dd9ee52bcf88649f60d58f20c81c0763e849/packages/cientos/src/core/controls/PointerLockControls.vue#L85).

### Steps to reproduce

- Add event listener (for example `isLock`) to the controls which logs a console message
- click the canvas multiple times
- see that the listener triggered too many times

### System Info

```shell

```

### Used Package Manager

pnpm

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/Tresjs/tres/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/Tresjs/.github/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://tresjs.org/guide).
- [x] Check that there isn't [already an issue](https://github.com/tresjs/tres/issues) that reports the same bug to avoid creating a duplicate.
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #1333, opened 2026-02-09
fix(cientos): some controls do not work in on demand render mode

Tinoooo:


packages/cientos/src/core/controls/PointerLockControls.vue
Tinoooo, line 1: **fyi:** I did not refactor this or fix [a related bug](https://github.com/Tresjs/tres/issues/1332) as the component needs an overhaul anyways and because it is too far out of the scope of this PR. Maybe it can be improved in #1332.

Tinoooo:
> LGTM, I like that you extracted similar props to `useOrbitLikeControls.ts`
> 
> Shouldnt we mention `screenSpacePanning` on the MapControls docs as well?

No, most props are not mentioned as there is the comment `All the props of the orbit controls component apply here too.`. I wouldn't make this inconsistent now.

PR #1335, opened 2026-02-12
feat(postprocessing): 1334 creation of effect pass components via custom agents

### LINKED ISSUE #1334: Creation of effect pass components via custom agents
### Description

As a developer of the postprocessing module of TresJS I want custom agents that help me create effect pass components.

### Suggested solution

create custom subagents

### Alternative

_No response_

### Additional context

_No response_

### Validations

Tinoooo:
fyi: this has the feat prefix because new effects will be a side-product of testing the agents.

closes TRES-273


packages/postprocessing/.claude/agents/component-suite-creator.md
```diff
@@ -0,0 +1,103 @@
...
... node_modules/postprocessing/src/effects/BloomEffect.js as a TresJS component?"\nassistant: "Absolutely! I'll use the vue-threejs-component-orchestrator agent to create a complete Vue component suite for the BloomEffect, including analysis, component generation, and optionally demos and playground examples."\n<commentary>\nThe user has provided a specific effect class path, so launch the vue-threejs-component-orchestrator agent to handle the complete workflow from analysis through component creation.\n</commentary>\n</example>\n\n<example>\nContext: User wants to add a new effect to the post-processing package.\nuser: "We need to add the TiltShift effect to @tresjs/post-processing"\nassistant: "I'll orchestrate the creation of the TiltShift effect component for the post-processing package. Let me launch the vue-threejs-component-orchestrator agent to analyze the effect class, generate the Vue component, and set up any demos or playground examples you need."\n<commentary>\nThis is a request to add a new effect component to the ecosystem, which requires the full orchestration workflow including analysis, component creation, and potentially demos/playground.\n</commentary>\n</example>
```
alvarosabu, line 3: Blocker:

The descriptions are oversized (wrapped by tge `<example><context>…<user>…<assistant>…<commentary>…</example>` blocks that are autogenerated by `/agents`) This pattern is used to help the dispatcher when they are many overlapping agents but I believe your intention is that they are dispatched by name from the orchestator.

These descriptions get loaded into the parent context everyutime the dispatcher evaluates, thats costly

packages/postprocessing/.claude/agents/component-suite-creator.md
```diff
@@ -0,0 +1,103 @@
...
+1. Analyze the Vignette effect class to understand its properties
+2. Generate a Vue component following TresJS conventions
+3. Optionally create a demo page and playground example
+
+First, could you confirm the path to the effect class? I'm assuming it's from pmndrs/postprocessing - is that correct?
+
+[After confirmation, proceed through each step, keeping the user informed]
+```
+
+Remember: Your goal is to make the component creation process smooth and educational. The user should understand what was created and feel confident about the next steps.
```
alvarosabu, line 103: Non-Blocking but important:

i believe this agent is rather a skill since it needs to interact with the user directly. Agents need to bubble it through task boundary mking it an agent means the dispatcher has to roundtrip every clarification.

packages/postprocessing/.claude/agents/three-effect-component-creator.md
alvarosabu, line 1: Non-Blocking but important:

This one is textbook definition of a skill. It's pure codegen from a JSON spec.

packages/postprocessing/.claude/agents/component-suite-creator.md
```diff
@@ -0,0 +1,103 @@
...
+- Share a summary of what was discovered with the user
+
+### Step 3: Generate the Vue Component
+- Use the `three-component-creator` sub-agent to generate:
+  - The Vue component with proper props
+  - TypeScript types
+  - Composables if needed
+  - Following TresJS conventions (see project CLAUDE.md files)
+- Show the user what was created and where files were placed
+
```
alvarosabu, line 38: @Tinoooo these two steps are sequential (component creator needs the output of the analyzer). Sub agents are mostly used for paralellism, the only subagents here that can be parallel are:

playground-creator edits apps/playground/src/router/routes/postprocessing/index.ts
docs-writer edits apps/postprocessing-docs-vitepress/.vitepress/config.ts

And they both call `AskQuestion` 

playground-creator Step 2: "You MUST use the AskQuestion tool to present your scene idea"
docs-writer Step 2.2: "You MUST use the AskQuestion tool to present your scene proposal"

Fire both in parallel → user gets two scene-design prompts arriving at unpredictable times for two different files, has to context-switch between them

alvarosabu (changes requested):
Hey @Tinoooo thanks for tackling this, I love that you went the dogfooding approach to validate the agentic tooling. I left some comments for specific issues:

 Two things worth a higher-level conversation: (1) the architecture should probably be skills + one agent rather than five agents, four of the five need `AskUserQuestion` mid-flow and the Step 5 "parallel dispatch" can't actually run parallel because the user becomes the serialization point

 (2) the new effect components have three correctness bugs (shared useRouteDisposal typing, v-for ref leak in SAOThreeDemo, wrong saoIntensity default) that are fixable but I think is better we do it once we merge the Nuxt docs and make the changes here.

let me know if you have any question

PR #1345, opened 2026-02-19
feat(core): slim createTresApp alternative for tree-shakeable Three.js usage

### LINKED ISSUE #1292: Tree-shakable Three.js Imports for TresJS
### Description

Currently, TresJS imports the entire Three.js namespace in `Context.vue` and registers all exports. This causes bundlers to include the **entire** Three.js library in the final build, even when the user only uses a small subset of Three.js objects.

For a typical start template provided by `pnpm create tres` with build config below:

```ts
  build: {
    minify: false,
    rolldownOptions: {
      output: {
        codeSplitting: {
          groups: [
            {
              test: /\/three\//,
              name: 'three',
            },
          ],
        },
      },
    },
  },
```

```shell
vite v8.0.0-beta.11 building client environment for production...
✓ 43 modules transformed.
dist/index.html                               0.63 kB │ gzip:   0.36 kB
dist/assets/index-eDIznQ1E.css                0.13 kB │ gzip:   0.11 kB
dist/assets/rolldown-runtime-2YtBIAaz.js      0.31 kB │ gzip:   0.22 kB
dist/assets/index-DZHj7OJD.js               365.31 kB │ gzip:  88.65 kB
dist/assets/three-BurP11D-.js             1,152.04 kB │ gzip: 221.15 kB
```

If I remove default `extend(THREE)` and manually register Three.js objects 
```ts
  extend({
    PerspectiveCamera,
    AmbientLight,
    Mesh,
    BoxGeometry,
    MeshNormalMaterial,
    DirectionalLight,
    AxesHelper,
    GridHelper,
  })
```

```shell
vite v8.0.0-beta.11 building client environment for production...
✓ 43 modules transformed.
dist/index.html                   0.54 kB │ gzip:   0.33 kB
dist/assets/index-eDIznQ1E.css    0.13 kB │ gzip:   0.11 kB
dist/assets/index-Cg3u0u30.js   365.15 kB │ gzip:  88.59 kB
dist/assets/three-DQ5wCXSo.js   783.91 kB │ gzip: 152.88 kB
```

221.15kB -> 152.88kB is quite BIG!

### Suggested solution

Option 1: add a new prop `manualExtend` to `TresCanvas` that allows users to opt-out of automatic Three.js namespace registration.

Option 2: provide a plugin that analyzes which Three.js classes are used and auto-imports & tree-shakes accordingly. This might be framework-specific.

### Alternative

_No response_

### Additional context

https://github.com/pmndrs/react-three-babel

### Validations

alvarosabu:
## Summary

- Add `createTresApp` programmatic API as a slim alternative to `<TresCanvas>` that does **not** auto-extend the full THREE namespace, enabling proper tree-shaking
- Add `slim.ts` entry point (`@tresjs/core/slim`) exporting only `createTresApp`
- Move `extend(THREE)` into `TresCanvas` directly (no change for existing users)
- Lazy-load devtools registration in `Context` (dev-only, dynamic import)
- Add `apps/bundle-test` app with Rollup Visualizer to compare full vs slim build sizes
- Add playground demo page for the manual-extend approach

## Usage

```ts
import { createTresApp } from '@tresjs/core/slim'
import { Mesh, BoxGeometry, MeshStandardMaterial } from 'three'
import MyScene from './MyScene.vue'

const app = createTresApp(canvas, { windowSize: true })
app.extend({ Mesh, BoxGeometry, MeshStandardMaterial })
app.render(MyScene)
```

## Test plan

- [ ] `<TresCanvas>` still works as before (extends all of THREE automatically)
- [ ] `createTresApp` renders scene correctly with only registered classes
- [ ] `createTresApp` with no `extend()` call does NOT include unused Three.js classes in bundle
- [ ] Existing tests pass (`pnpm t`)
- [ ] Bundle-test app builds and visualizer shows size difference between full/slim modes

Closes #1292 
Closes TRES-214


packages/core/src/components/TresScene.test/extends.test.ts
```diff
@@ -0,0 +1,40 @@
...
+  return { sceneWrapper, context, catalogue }
+}
+
+describe('tresCanvas - catalogue', () => {
+  it('extends all of THREE by default', async () => {
+    const { catalogue, sceneWrapper } = await createScene()
+    expect(catalogue.value).toHaveProperty('Mesh')
+    expect(catalogue.value).toHaveProperty('BoxGeometry')
+    sceneWrapper.unmount()
+  })
```
Tinoooo, line 39: **suggestion(non-blocking):** It would be nice to have a test for the manual extending here.

packages/core/src/composables/createTresApp/index.ts
```diff
@@ -0,0 +1,136 @@
+import type { Component, ShallowRef } from 'vue'
+import { ACESFilmicToneMapping, PCFSoftShadowMap } from 'three'
+import { createApp, defineComponent, h, shallowRef } from 'vue'
+import type { ContextEmits, ContextProps } from '../../components/Context.vue'
+import Context from '../../components/Context.vue'
+import { extend as catalogueExtend } from '../../core/catalogue'
+import type { TresContext } from '../useTresContextProvider'
+
+type ContextExposed = { context: ShallowRef<TresContext>, dispose: () => void }
```
Tinoooo, line 9: **suggestion:** The lint check fails. This should be an interface.

```suggestion
interface ContextExposed { context: ShallowRef<TresContext>, dispose: () => void }
```

packages/core/src/composables/createTresApp/index.ts
```diff
@@ -0,0 +1,136 @@
...
+  Object.assign(canvas.style, {
+    display: 'block',
+    width: '100%',
+    height: '100%',
+    position: windowSize ? 'fixed' : 'relative',
+    top: '0',
+    left: '0',
+    pointerEvents: 'auto',
+    touchAction: 'none',
+  })
```
Tinoooo, line 74: **suggestion:** For better maintainability, I suggest putting the styles in a central position and using them in both this composable and TresCanvas.

packages/core/src/composables/createTresApp/index.ts
```diff
@@ -0,0 +1,136 @@
...
+
+  const slotComponent = shallowRef<Component | null>(null)
+  let contextInstance: ContextExposed | null = null
+  const eventHandlers: Partial<Record<keyof ContextEmits, ((...args: unknown[]) => void)[]>> = {}
+
+  function buildEmitBindings(): Record<string, (...args: unknown[]) => void> {
```
Tinoooo, line 80: **nitpick:** I think arrow functions are more common in our codebase.


```suggestion
  const buildEmitBindings = (): Record<string, (...args: unknown[]) => void> => {
```

packages/core/src/composables/createTresApp/index.ts
```diff
@@ -0,0 +1,136 @@
...
+    get context() {
+      return contextInstance?.context?.value
+    },
+    on(event, handler) {
+      if (!eventHandlers[event]) {
+        eventHandlers[event] = []
+      }
+      eventHandlers[event]!.push(handler as (...args: unknown[]) => void)
+      return tresApp
+    },
```
Tinoooo, line 132: **question:** How is this supposed to work? `eventHandlers` is already used above in it's original state. I tried using event hooks for context emits here, but it dod not work. What was the intention here?

apps/bundle-test/src/main-slim.ts
```diff
@@ -0,0 +1,22 @@
...
+import SceneCreateTres from './SceneCreateTres.vue'
+
+const canvas = document.getElementById('canvas') as HTMLCanvasElement
+
+const root = createTresApp(canvas, { windowSize: true })
+
+root
+  .extend({ AmbientLight, BoxGeometry, DirectionalLight, Mesh, MeshStandardMaterial, PerspectiveCamera })
+  .render(SceneCreateTres)
+ */
```
Tinoooo, line 19:
```suggestion
```

apps/bundle-test/dist-tsdown/slim.js
Tinoooo, line 1: **issue/suggestion/question:** The bundle test pages throw an error and don't show anything. Shall we maybe not have the `bundle-test` folder in the repo? This might reduce maintenance work in the future.

PR #1342, opened 2026-02-15
docs(rapier): nuxt rapier docs

JaimeTorrealba:
closes #1303

Fixes TRES-208

PR #1395, opened 2026-03-27
feat(rapier): contact force

### LINKED ISSUE #1313: Contact force events
*Migrated from Tresjs/rapier#106 - Originally created by @JaimeTorrealba*

---

### Description

You should be able to emit the contact force events, when 2 objects collide

### Suggested solution

Could be base on: https://github.com/pmndrs/react-three-rapier?tab=readme-ov-file#contact-force-events Using Vue Interface for emit and events

### Alternative

_No response_

### Additional context

_No response_

### Validations

- [X] I agree to follow this project's [Code of Conduct](https://github.com/Tresjs/rapier/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/Tresjs/rapier/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://rapier.tresjs.org/guide).
- [X] Check that there isn't [already an issue](https://github.com/tresjs/rapier/issues) that reports the same bug to avoid creating a duplicate.

JaimeTorrealba:
closes #1313

Closes TRES-235


apps/playground/src/pages/rapier/ContactForceDemo.vue
```diff
@@ -0,0 +1,106 @@
...
+      <Physics :debug>
+        <!-- Falling sphere with contact force events enabled -->
+        <RigidBody
+          ref="sphereRef"
+          :active-contact-force="activeContactForce"
+          :contact-force-event-threshold="FORCE_THRESHOLD"
+          @contact-force="onContactForce"
+          collider="ball"
+        >
+          <
```
alvarosabu: Blocking:

You have a stray `<` here.

apps/playground/src/pages/rapier/ContactForceDemo.vue
```diff
@@ -0,0 +1,106 @@
...
+const sphereRef = shallowRef();
+
+const onContactForce = (event: any) => {
+  const magnitude = Math.round(event.totalForceMagnitude);
+  console.log("jaime ~ onContactForce ~ magnitude:", magnitude);
+};
+const onContactForceCustomCollider = (event: any) => {
+  const magnitude = Math.round(event.totalForceMagnitude);
+  console.log("jaime ~ onContactForceCustomCollider ~ magnitude:", magnitude);
+};
```
alvarosabu: Blocker:

Avoid adding author's name to the logs.

apps/playground/src/pages/rapier/ContactForceDemo.vue
```diff
@@ -0,0 +1,106 @@
...
+  toneMapping: ACESFilmicToneMapping,
+};
+
+const sphereRef = shallowRef();
+
+const onContactForce = (event: any) => {
+  const magnitude = Math.round(event.totalForceMagnitude);
+  console.log("jaime ~ onContactForce ~ magnitude:", magnitude);
+};
+const onContactForceCustomCollider = (event: any) => {
```
alvarosabu: Suggestion:

We will get more strict with types on the playgrounds in near future, let's start by typing these events correctly.

alvarosabu: @JaimeTorrealba I think you can use `ContactForcePayload`

packages/rapier/src/utils/collisions.ts
```diff
@@ -44,3 +44,15 @@ export const collisionTrigger = (
...
+
+export const contactForceTrigger = (
+  source: SourceTarget,
+  target: SourceTarget,
+  payload: Omit<ContactForcePayload, 'source' | 'target'>,
+) => {
+  const [rigidBodyObject, colliderObject] = source.objects
+  const eventPayload: ContactForcePayload = { source, target, ...payload }
+
+  rigidBodyObject?.__vnode?.ctx?.emit?.('contact-force', eventPayload)
```
alvarosabu, line 56: Question:

@JaimeTorrealba is it normal to log so many similar events? 

Image

alvarosabu (changes requested):
Hi @JaimeTorrealba thanks a lot for this, looks pretty great already. I left some comments before we can merge.

Maybe is worth to merge #1342 first so you can add the docs to this one?

packages/rapier/src/components/RigidBody.vue
```diff
@@ -137,6 +139,9 @@ watch(() => props.activeCollision, value => setAutoColliderProp('activeCollision
 watch(() => props.activeCollisionTypes, value => setAutoColliderProp('activeCollisionTypes', value))
 watch(() => props.collisionGroups, value => setAutoColliderProp('collisionGroups', value))
 watch(() => props.sensor, value => setAutoColliderProp('sensor', value))
+watch(() => props.activeContactForce, value => setAutoColliderProp('activeContactForce', value))
+watch(() => props.contactForceEventThreshold, value => setAutoColliderProp('contactForceEventThreshold', value))
```
alvarosabu, line 144: Probably better to set the contactForce threeshold so something meaningful, there is a lot of noise in that console log in rest position

packages/rapier/src/utils/collisions.ts
JaimeTorrealba: Yes this is perfectly normal (I also have the doubt) that's why it uses a threshold. I'll explain it on the docs

apps/playground/src/pages/rapier/ContactForceDemo.vue
alvarosabu, line 1: Blocking:

@JaimeTorrealba I don't quite get how the demo is supposed to work  

https://github.com/user-attachments/assets/2e09fbde-38f5-4aea-b43b-0a7a4daa01ef

alvarosabu (changes requested):
Hi @JaimeTorrealba thanks for adding the docs, I'm not clear how the demos and the playground suppose to work, am I missing something?

apps/playground/src/pages/rapier/ContactForceDemo.vue
JaimeTorrealba: You need to build again, that happened to me too @alvarosabu

alvarosabu: @JaimeTorrealba what should I build, the rapier package?

JaimeTorrealba: On the root directory just run pnpm build (this build all, but it works)

packages/rapier/src/components/RigidBody.vue
JaimeTorrealba: I went for a default of `5.0`  which I found a reasonable number remove most of the logs, I think going high could potentially confuse users if the event is not triggered when the contact is close enough (the default on rapier is 0)

Let me know what you think

Neosoulink: If I'm not wrong, it's normal to have such an amount of logs in our case, based on the default **gravity** applied and the **threshold**. 

We can either reduce the gravity or increase the **threshold** to 15 to have fewer logs.

apps/playground/src/pages/rapier/ContactForceDemo.vue
Neosoulink: It's working on my side, but I can suggest not triggering logs in the documentation; instead, we can add more messages at the top-left, like what you did for the `Total force magnitude: *`

Neosoulink (approved):
This is an amazing integration as usual, @JaimeTorrealba !!!🚀🔥

> I just left a minor comment here: [#r3074266143](https://github.com/Tresjs/tres/pull/1395/changes#r3074266143)

apps/playground/src/pages/rapier/ContactForceDemo.vue
JaimeTorrealba: unmm I kind of dislike having logs, but honestly I don't have any other idea how to show the list of event (that can be very big) into the screen without overwhelm the UI, any ideas here @alvarosabu ?

alvarosabu: Hey @JaimeTorrealba so I run install and build rapier and now I can see the logs 👍🏻 thanks!

I think console logs are fine since the user can use the devtools console to inspect the the contact force object, it would be cumbersome to do this on the UI. @Neosoulink

alvarosabu (approved):
LGTM @JaimeTorrealba great job 👏

PR #1400, opened 2026-04-11
fix(core): reuse custom renderer per canvas to stop HMR duplicates

### LINKED ISSUE #23: HMR disposal
Atm HMR disposal is not working as expected. Meshes are duplicated and this could lead to performance issues on large scenes.

Also not nice for DX. User needs to reload everytime.

This one is pretty tricky.

# Desired solution.

Renderer, scene and object follow a disposal flow when HMR hits

alvarosabu:
## Summary

Fixes HMR duplicating scene objects on template edits.

**Root cause:** `Context.vue#mountCustomRenderer` was creating a fresh Vue custom renderer on every `vite:afterUpdate` event. A fresh renderer has no prior vnode tree, so Vue's diff runs against `null` and mounts everything as new while the old scene objects remain parented to `scene` — duplicates accumulate every HMR tick.

**Fix (R3F-inspired):** One Vue custom renderer per canvas, stored in a module-scoped `_roots: Map<HTMLCanvasElement, TresRoot>`. The map survives HMR because module state persists. On `vite:afterUpdate`, instead of rebuilding, a reactive `hmrTick` ref is bumped — the internal component's render function tracks it as a dep and re-runs, re-reading `slots.default()`. Vue's custom renderer then diffs old vs new and calls `nodeOps.remove` / `patchProp` / `insert` to sync the Three.js scene graph in place.

### What changed

- **New** `packages/core/src/core/roots.ts` — module-scoped per-canvas renderer state (`getRoot` / `setRoot` / `deleteRoot` / `hasRoot` + `TresRoot` type)
- `Context.vue#mountCustomRenderer` is now idempotent — lazily creates one renderer per canvas, reuses on subsequent calls
- `Context.vue#createInternalComponent` takes an `hmrTick: Ref<number>` and reads it reactively inside the render function
- `Context.vue#handleHMR` → `getRoot(canvas).hmrTick.value++`
- `Context.vue#unmountCanvas` uses `root.renderer.render(null, scene)` so Vue walks the vnode tree and disposes via `nodeOps.remove`, then `deleteRoot(canvas)`
- New `onBeforeUnmount` cleans up the `vite:afterUpdate` listener (also fixes a pre-existing leak)
- `registerTresDevtools` now runs exactly once per mount instead of every HMR tick (pre-existing duplicate-registration bug quietly fixed)

### Tests added

`packages/core/src/components/TresScene.test/hmr.test.ts` (8 tests + 1 documented `it.skip`):
- No duplication across simulated HMR tick
- Slot content diff with a non-reactive mutable (pins `hmrTick` as the load-bearing trigger)
- Root entry deleted on unmount
- Primitive object identity preserved across HMR (`<primitive :object="mesh">` keeps the same THREE reference)
- `registerTresDevtools` called exactly once per mount
- Empty default slot + undefined default slot both mount cleanly
- `it.skip` for `vite:afterUpdate` listener cleanup — mocking `import.meta.hot` doesn't reach `Context.vue` across `vi.resetModules()`; cleanup is verified by inspection and documented in the test comment

Plus `packages/core/src/core/roots.test.ts` with 5 unit tests for the map helpers in isolation.

Full core suite: 528 passed + 3 skipped.

## Breaking changes

**None on public API.** `TresCanvasInstance` shape is unchanged. `TresCanvasInstance.dispose()` runtime behavior is preserved — the `force` parameter on the internal `dispose` helper was restored after a late review caught that the manual-dispose path had silently lost its WebGL teardown.

**Behavioral note worth calling out in release notes:** HMR no longer wipes imperative scene state that was previously getting destroyed on every template edit. Most users will experience this as an improvement (OrbitControls target survives, primitive GLTFs don't re-download). Edge-case code that relied on the implicit reset may behave differently.

## Ecosystem impact

| Package | Risk | Notes |
|---|---|---|
| `@tresjs/cientos` | Low | Uses composables, not mount/unmount internals. Worth a smoke test on OrbitControls/CameraControls + loaders. |
| `@tresjs/postprocessing` | Low-medium | `<EffectComposer>` wraps the WebGL render loop. Worth verifying effect pass HMR. |
| `@tresjs/nuxt` | Medium (SSR) | `_roots` Map lives at module level. Nuxt's vite-node SSR runtime isolates modules per request, so it should be fine, but warrants an explicit SSR test. |
| `@tresjs/rapier` | Low | Physics composables, no mount/unmount touch. |
| `@tresjs/leches` | None | Standalone. |

## Test plan

- [ ] Unit tests pass (`pnpm --filter @tresjs/core test`)
- [ ] Lint + build clean (`pnpm --filter @tresjs/core lint` / `build`)
- [ ] Manual HMR verification at `/issues/23-hmr-disposal`:
  - [ ] Baseline `scene.children.length` counter stable
  - [ ] Edit template: add `<TresMesh><TresSphereGeometry /></TresMesh>` → counter goes up by exactly 1, no duplicates
  - [ ] Edit template: remove it → counter back to baseline
  - [ ] Rotate with OrbitControls, then edit template → camera angle survives
  - [ ] Edit `Model.vue` script → no duplicates
  - [ ] DevTools → inspect `scene.children` → no orphans
- [ ] Smoke test cientos controls/loaders against the new HMR flow
- [ ] `@tresjs/nuxt` SSR sanity check

## Follow-ups (not this PR)

- Replace `it.skip` with an integration test via a `src/core/viteHot.ts` shim
- Consider renaming `dispose` → `disposeSceneGraph` or extracting `mountCustomRenderer` into `createTresRoot.ts`
- Verify Nuxt SSR safety with an explicit test in `@tresjs/nuxt`

Closes #23


packages/core/src/components/Context.vue
```diff
@@ -127,16 +129,34 @@ const createInternalComponent = (context: TresContext, empty = false) =>
...
 
-const mountCustomRenderer = (context: TresContext, empty = false) => {
-  const InternalComponent = createInternalComponent(context, empty)
-  const { render } = createRenderer(nodeOps({ context, options: props.customRendererOptions }))
-  render(h(InternalComponent), scene.value as unknown as TresObject)
+const mountCustomRenderer = (context: TresContext) => {
+  const canvas = props.canvas
+  if (getRoot(canvas)) { return }
+
+  const hmrTick = shallowRef(0)
```
Tinoooo:
```suggestion
  const hmrTick = ref(0)
```

Tinoooo (dismissed):

packages/core/src/components/Context.vue
alvarosabu: Done @Tinoooo

PR #1411, opened 2026-05-25
feat(cientos): add `<Decal>` abstraction with interactive editor UI

damienmontastier:
# Decal
  This component provides an abstraction over **`DecalGeometry`** from `three-stdlib`,
  `<Decal />` projects a **flat texture onto the surface of a parent mesh**, conforming to
  its geometry. It supports *stacking* via per-mesh `zIndex`, *lossless JSON serialization*
  via `v-model:data`, and an optional **interactive editor overlay** for authoring layouts
  directly in the canvas.
  - Drop-in **`<DecalDebugUI />`** overlay — handle, dock, layer panel, *no host code to 
  wire*
  - **Programmatic API** via `useDecalEditor()` for custom panels and automation
  - **Multiple textures palette** — pass an array of `Texture` to `:map`
  - **Custom material** override via the default slot
  - **GLB / loaded models** — `:mesh` prop targets a non-parent mesh (auto re-parents so the
  decal follows runtime transforms)
  - **BVH-accelerated** projection — auto-detected via `useBVH`, 10–100× faster on dense
  meshes
  - **Per-mesh stacking** with automatic polygon-offset
  - **Per-decal tint** (color + opacity), **mirror** (flip X/Y), **visibility toggle**,
  **undo / redo**
  - **Import / Export** JSON (auto-named download + file-picker import with sanitisation)
  - **Keyboard shortcuts** — `Enter` commit, `Esc` cancel, `Del` delete, `⌘Z` / `⇧⌘Z`
  undo/redo
  ### Notes / Feedback
  - `editable` prop emits a `console.warn` to remind users the interactive editor is for
  *authoring*, not *production* — strip it (and unmount `<DecalDebugUI />`) before shipping;
  persist the resulting `v-model:data` JSON instead.
  - `DecalGeometry` baking is **static**: `SkinnedMesh` skinning, morph targets, and direct
  mutations of `geometry.attributes.position` on the parent are **not** followed.
  - **Wrap-around bug** ([[three.js#21187](https://github.com/mrdoob/three.js/issues/21187)](https://github.com/mrdoob/three.js/issues/21187))
  mitigated by the `cullThreshold` prop (default `0.2`). Complex / thin-leaf meshes (plants,
  fabrics) may need a tighter value like `0.7`.


packages/cientos/src/core/abstractions/Decal/ROADMAP.md
alvarosabu, line 1: Blocking:

@damienmontastier The intention of this file is awesome, but I think it should live inside the docs instead and we should probably create github tickets out of this for pending tasks, wdyt?

packages/cientos/src/core/abstractions/Decal/DebugUI/DecalDebugUI.vue
```diff
@@ -0,0 +1,244 @@
...
+// `opacity: 0` defaults apply on the first paint — otherwise the handle
+// flashes for one frame at its initial position. Self-contained: no
+// `import 'pkg/style.css'` boilerplate, no exports-map plumbing.
+const STYLE_ID = 'cientos-decal-ui-styles'
+if (typeof document !== 'undefined' && !document.getElementById(STYLE_ID)) {
+  const tag = document.createElement('style')
+  tag.id = STYLE_ID
+  tag.textContent = DECAL_DEBUG_UI_STYLES
+  document.head.appendChild(tag)
+}
```
alvarosabu: Blocking:

I understand the rationale @damienmontastier but the cost is high, is 975 lines of CSS as inlined JS string  bloating the bundle for all cientos consumers  even those not using <DecalDebugUI>. Tree-shakable in theory (only imported by that component), but the string is non-gzippable separately and not purgeable.

+ It breaks under strict CSP (style-src 'self' without unsafe-inline blocks `document.createElement('style')` injection).

Some options

- `<style>` (not scoped) inside the SFC + accept the one-time CSS-import step in docs
- CSS Modules with sync injection via Vite's `?inline` query
- Inline critical styles for the initial-paint elements (just the opacity: 0 rule), keep the rest in the style tag

apps/cientos-docs/app/components/abstractions/Decal.vue
```diff
@@ -0,0 +1,122 @@
...
+
+watch(textures, (list) => {
+  if (Array.isArray(list)) {
+    list.forEach((t) => { if (t) { t.colorSpace = SRGBColorSpace } })
+  }
+}, { immediate: true })
+
+const { state: model } = useLoader(
+  GLTFLoader,
+  'https://cdn.jsdelivr.net/gh/Tresjs/assets@main/models/gltf/kuwahara-effect/plant-jar/plant-jar.glb',
```
alvarosabu: Suggestion:

I think I would rather have a mug, like the one used here https://lab.tresjs.org/experiments/coffee-smoke, I can push one white if you want. A metal one with white cover would be cool look 

Image

alvarosabu (changes requested):
Hey @damienmontastier thanks a lot of taking the time for tackling this. I believe is a very important missing abstraction and users would love to have it.

I left a couple of blockers before we can merge, let me know if you have any question

PD: Big kudos for the amazing Decal Editor UX, it's beautiful

packages/cientos/src/core/abstractions/Decal/DebugUI/DecalDebugUI.vue
damienmontastier: Good call @alvarosabu  — went with option 1 (CSS file imported by the SFC, not a `<style>` block, so tsdown extracts it cleanly to a single `dist/trescientos.css`).

Changes:
  - `styles.ts` (975-line JS string) → `styles.css` imported from `DecalDebugUI.vue`. Bundle drops ~22 KB of JS, CSS file gzips to ~4 KB and is purgeable.
  - No more `document.createElement('style')` injection — CSP-safe (`style-src 'self'` no longer breaks it).
  - `package.json` exposes `./styles.css` → `./dist/trescientos.css` with `sideEffects: ["**/*.css"]` so the import isn't tree-shaken away.
  - Docs (`decal.md`) ship a `prose-warning` with the one-time import snippet for both Vite (`main.ts`) and Nuxt (`nuxt.config.ts`). `<Decal>` itself stays style-less — the import is only needed when mounting `<DecalDebugUI>`.

Let me know if you'd prefer a different export name than `./styles.css` (kept generic so future cientos components can reuse it).

alvarosabu (approved):
LGTM @damienmontastier thanks a lot for this. Its gonna be massive

PR #1403, opened 2026-04-18
feat(rapier): full collider shapes support

Neosoulink:
# Summary

- `RigidBody` component now uses `RigidBodyCollidersShape` via the `collider` prop to generate auto colliders
  - Supported shapes: `cuboid`, `ball`, `capsule`, `cone`, `cylinder`, `convexHull`, `trimesh`.
- Full Collider Shapes support
  - Supported `ColliderShape`: `ball`, `capsule`, `cone`, `convexHull`, `convexMesh`, `cuboid`, `cylinder`, `heightfield`, `polyline`, `roundCone`, `roundConvexHull`, `roundConvexMesh`, `roundCuboid`, `roundCylinder`, `trimesh`.
- Align `Collider` components with `ColliderShape`.
- Code clean-up and accessibility improvements.

## Breaking changes

This integration was designed to be backwards compatible, but some changes were necessary to align with the new `ColliderShape`.

- Old `hull` collider no longer supported, use `convexHull` instead.

PR #1419, opened 2026-05-28
fix(rapier): collision events & collider args resolution

Neosoulink:
# Summary

Fixes inconsistent collision event dispatch and improves collider argument typing to match Rapier’s ColliderDesc signatures.

### Event dispatch

* Contact-force events (in [Physics.vue](<https://github.com/Tresjs/tres/blob/main/packages/rapier/src/components/Physics.vue#L90-L93>))
  * Now use the right `getNodeObjectsFromCollisionSource`.

    > This might be caused by a merge conflict or something like that, since it was introduced here Tresjs/tres#1403
* Sensor intersections (in [intersections.ts](<https://github.com/Tresjs/tres/blob/main/packages/rapier/src/utils/intersections.ts#L13-L38>))
  * Now read the collider object from `target.objects`.
  * Emit intersection events on both `RigidBody` and `Collider` components when their sensor prop is set.
* Sync body type, auto-collider shape, collider geometry, transforms, and<br>material/event props with Rapier when component props change after mount.

### Collider typing

* All shape collider components (`Ball`, `Capsule`, `Cone`, etc...)  are now **Rapier-derived** using `ColliderDesc.{shape}` for accurate typings.
  * Also fixed incorrect collider args in the playground.


packages/rapier/src/components/colliders/Base.vue
```diff
@@ -69,26 +84,60 @@ watch(bodyContext, async (state) => {
...
-    'mass',
-    'activeCollisionTypes',
-    'activeCollision',
-    'sensor',
-  ],
-  colliderInfos,
-)
-
+// Props watchers
+watch(() => props.shape, value => {
```
alvarosabu: Blocking:

I think the new reactivity ended up layered on top of the old one rather than replacing it. We've still got `makePropsWatcherCL(props, ['friction', 'restitution', 'density', 'mass', ...], colliderInfos)` running, and now individual `watch(() => props.friction, ...), watch(() => props.mass, ...)`, etc. So each of those props fires its setter twice on every change.

The bigger one is `activeCollision / activeContactForce`. They each got their own watcher that calls `setActiveEvents `with just one flag, so toggling one wipes the other, enable contact force and you silently kill collision events, and vice versa. The existing combined `watch([() => props.activeCollision, () => props.activeContactForce, colliderInfos], ...)` already handles this correctly by OR-ing both flags together.

Could you pick one source of truth? Either keep `makePropsWatcherCL` + the combined active-events watcher and drop the new individual ones, or go fully individual and remove the batch watcher — but the two-flag active-events case needs to stay combined either way so they don't clobber each othe

packages/rapier/src/components/colliders/Base.vue
```diff
@@ -69,26 +84,60 @@ watch(bodyContext, async (state) => {
...
+watch(() => props.args, value => {
+  if (!colliderInfos.value?.collider) { return }
+  colliderInfos.value.collider.setShape(
+    (ColliderDesc[props.shape](
+      value?.[0] ?? 1,
+      value?.[1] ?? 1,
+      value?.[2] ?? 1,
+      value?.[3] ?? 1,
+    ) ?? ColliderDesc.cuboid(1, 1, 1)).shape)
+})
```
alvarosabu: Blocking:

This hard-codes the assumption that every collider factory takes up to **4** positional numbers, each defaulting to 1. That's true for `ball/cuboid/capsule/cone/cylinder`, but false for the mesh-family shapes whose leading args are typed arrays:

- trimesh(vertices: Float32Array, indices: Uint32Array)
- convexHull(points: Float32Array)
- convexMesh(vertices, indices)
- polyline(vertices: Float32Array, indices?)
- heightfield(nrows, ncols, heights: Float32Array, scale: Vector3)

alvarosabu: The `?? ColliderDesc.cuboid(...)` won't catch this either, it's a nullish fallback, and these cases throw, so it propagates out of the watcher.

alvarosabu (changes requested):
Hi @Neosoulink thanks a lot for this PR, I'm afraid it needs more work, let me know if you have any questions

packages/rapier/src/components/colliders/Base.vue
Neosoulink: Yeah, I agree with the default arguments fallback issue, let me update that.

Regarding the Cuboid fallback, it's to catch `ColliderDesc[value]?.shape`, since the `Shape` can also be `Null`

packages/rapier/src/components/colliders/Base.vue
Neosoulink: Combining is much better since `makePropsWatcherCL` only has minimal logic to set the new property value

Neosoulink:
Hey @alvarosabu, Thanks for those clear reviews!
Effectively, they were duplications.

I've just pushed a fix here: https://github.com/Tresjs/tres/pull/1419/commits/e3b66de4e053d7ae045e1265487aa65f19bb34fe

I've tested on my side, and both Demos and Lab experimentation, and all's working on my end.

Neosoulink:
Hey @alvarosabu, I've rebased/updated this PR first since #1429 depends on this one ✅

Neosoulink:
Thanks for the updates @alvarosabu 🎉

PR #1430, opened 2026-06-09
chore(cientos): shift responsibility of preset quality validation to …

Tinoooo:


alvarosabu (approved):
@Tinoooo thanks for this, its better approach

PR #1440, opened 2026-06-19
feat(core): expose isWebGPURenderer guard + reactive isWebGPU flag

alvarosabu:
## Summary

Surfaces the renderer type to the ecosystem so components can branch internally. Core already supports WebGPU and had the detection logic internally — this promotes it to public API.

- Add public guards `isWebGPURenderer(value): value is Renderer` and `isWebGLRenderer(value): value is WebGLRenderer` in `utils/is/tres.ts`, following the existing `createTypeGuard` pattern. The `isRenderer` / `isWebGLRenderer` marker flags are mutually exclusive (verified against three 0.184), so the guards never both match.
- Expose a reactive `isWebGPU: ComputedRef<boolean>` on `TresContext`, computed from `renderer.instance`. Also surfaced through `useTres()`.
- Reuse the new `isWebGPURenderer` guard inside `useRendererManager` (drops the duplicated inline `isRenderer`).
- Unit tests for both guards, mirroring the existing guard suite.

### Notes

`WebGPURenderer` can run on a WebGPU **or** a WebGL2 fallback backend. The guard identifies the renderer *instance*, not the active backend — the right granularity here (`NodeMaterial` works on both backends; GLSL `ShaderMaterial` works on neither under the WebGPU renderer).

`three`'s `WebGLRenderer` type omits the runtime `isWebGLRenderer` flag, so `isWebGLRenderer` intersects `{ isWebGLRenderer: true }` to satisfy `createTypeGuard`'s `keyof`.

## Test plan

- [x] `import { isWebGPURenderer } from '@tresjs/core'` works (exported via main index, confirmed in `dist/tres.d.ts`)
- [x] `const { isWebGPU } = useTresContext()` is reactive and correct for both renderers
- [x] Unit tests pass (`632 passed | 3 skipped`)
- [x] Typecheck clean (no new errors)
- [x] Lint + build pass

Closes TRES-329

### DIFF
--- a/packages/core/src/composables/useRenderer/useRendererManager.ts
+++ b/packages/core/src/composables/useRenderer/useRendererManager.ts
@@ -19,7 +19,7 @@ import { logWarning } from '../../utils/logger'
 import type { SizesType } from '../useSizes'
 import type { UseCameraReturn } from '../useCamera'
 import type { TresScene } from '../../types'
-import { isFunction, isObject } from '../../utils/is'
+import { isFunction, isWebGPURenderer } from '../../utils/is'
 import { useCreateRafLoop } from '../useCreateRafLoop'
 import { TresRendererError } from '../../utils/error'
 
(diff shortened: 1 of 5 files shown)


alvarosabu:
@Tinoooo there is a point I would love to know your opinion about, if we should add the computed on the context or just the guard. let me know.

Tinoooo (approved):

PR #1442, opened 2026-06-20
fix(docs): restore MDC page slots mangled into headings

Gazoon007:
## Problem

  The three docs landing pages render broken and throw a Vue hydration mismatch on the homepage. #1437 bumped `@eslint/markdown` to v8, where `markdown/no-missing-atx-heading-space` is on by default. A `lint --fix` run then rewrote the MDC slot directives (`#title`, `#description`, `#features`) into ATX headings (`# title`, ...) in `content/index.md` across cientos-docs, docs-boilerplate and postprocessing-docs, plus the cientos-docs `CLAUDE.md` example.

  `remark-mdc` only treats `#name` as a slot when a letter follows the `#` immediately. With a space, it falls back to CommonMark and becomes an `<h1>`, so the hero/section/feature text stopped filling the Nuxt UI components' named slots and fell into their default slots. That is what broke the pages and triggered the hydration mismatch.

  ## Fix

  - Revert the slot syntax in the three `index.md` landing pages and the cientos-docs `CLAUDE.md` example (now byte-identical to the pre-#1437 state).
  - Turn off `markdown/no-missing-atx-heading-space` (and keep `no-multiple-h1` off) for `content/**/*.md`, so the linter stops rewriting MDC slot directives back into headings. Also corrected the misleading comment that called `# title` the MDC syntax.

  ## Verification
  
  `pnpm --filter cientos-docs lint && pnpm --filter docs-boilerplate lint && pnpm --filter postprocessing-docs lint` passes clean.

### DIFF
--- a/apps/cientos-docs/CLAUDE.md
+++ b/apps/cientos-docs/CLAUDE.md
@@ -202,10 +202,10 @@ cta:
   - /getting-started
 ---
 
-# title
+#title
 Hero Title
 
-# description
+#description
 Hero description
 ::
 ```
--- a/apps/cientos-docs/content/index.md
+++ b/apps/cientos-docs/content/index.md
@@ -23,12 +23,12 @@ Tres v5 announcement
 :::
  -->
 
-# title
+#title
 
 <span class="text-primary-300">Cientos</span> \
 Ready-made abstractions for TresJS
 
-# description
+#description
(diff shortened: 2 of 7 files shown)


alvarosabu (approved):
Thanks a lot @Gazoon007 this is clearly the right way

PR #1445, opened 2026-06-22
feat: MeshPortalMaterial (MVP) + TresPortal

alvarosabu:
## Summary

Adds portal support to TresJS in two layers:

- **`@tresjs/core` — `<TresPortal>`**: reparents its declarative children into any target `Object3D`/`Scene` via Vue `<Teleport>` (reactive add/remove, `disabled` prop to render in place).
- **`@tresjs/cientos` — `<MeshPortalMaterial>`**: renders its children into a private off-main scene and projects that scene onto the host mesh as a perspective-correct, geometry-masked window (port of [drei MeshPortalMaterial](https://drei.docs.pmnd.rs/portals/mesh-portal-material)).
  - Window pass renders the portal scene to an FBO each frame; the mesh samples it in screen space so contents sit in world space (camera movement → parallax through the surface).
  - `blend` cross-fades world (`0`) ↔ portal (`1`); `blend === 1` renders the portal scene fullscreen ("inside" the portal).
  - Overrides the injected scene context for its children so imperative helpers (`<Environment>`) and declarative `attach="background"` target the **portal** scene, not the world — enabling per-portal environments/backgrounds.

### Props (`MeshPortalMaterial`)

| Prop | Default | Description |
| --- | --- | --- |
| `blend` | `0` | `0` = world only, `1` = portal only; cross-fades between. |
| `resolution` | `512` | FBO resolution for the window pass. |
| `worldUnits` | `false` | Keep portal contents in world units instead of host-relative. |
| `renderPriority` | `0` | Render-loop priority for the blend takeover (applied at mount). |

## Known limitations (MVP, documented)

- `blur` / SDF edge-fade not yet implemented.
- No pointer-event forwarding into the portal scene.
- WebGPU not yet supported (warns once, no-ops).
- Intermediate `blend` (0<blend<1) does a fullscreen cross-fade rather than a geometry-masked transition; best used as a 0↔1 toggle.

## Test plan

- [x] `pnpm --filter @tresjs/core test run portal` — 4 tests passing (reparenting, reactive add/remove, scene-context override)
- [x] Portal source typechecks clean (vue-tsc)
- [x] Playground demo `cientos/materials/portal-material` renders the 3-portal gallery with per-portal environments and blend slider

## Docs

- cientos: `Mesh Portal Material` API page with live demo
- core: `TresPortal` component page with example


packages/core/src/components/TresPortal.vue
```diff
@@ -0,0 +1,16 @@
...
+defineProps<{
+  /** Target Object3D/Scene to reparent children into. */
+  to: Object3D
+  /** When true, children render in place (main scene) instead of the target. */
+  disabled?: boolean
+}>()
+</script>
+
+<template>
+  <Teleport :to="(to as any)" :disabled="disabled">
```
JaimeTorrealba: is ok to have an any here?

packages/cientos/src/core/materials/meshPortalMaterial/PortalMaterialImpl.ts
```diff
@@ -0,0 +1,49 @@
...
+      uniforms: {
+        map: { value: null },
+        blend: { value: 0 },
+        resolution: { value: new Vector2() },
+      },
+      vertexShader: /* glsl */`
+        varying vec2 vUv;
+        void main() {
+          gl_Position = projectionMatrix * modelViewMatrix * vec4(position, 1.0);
+          vUv = uv;
```
JaimeTorrealba: We're no using the vUv, I see you mention that in the future we plan to use it (with the packing include) but I believe we can comment them for now

packages/cientos/src/core/materials/meshPortalMaterial/PortalMaterialImpl.ts
```diff
@@ -0,0 +1,49 @@
...
+        }
+      `,
+      fragmentShader: /* glsl */`
+        uniform sampler2D map;
+        uniform float blend;
```
JaimeTorrealba: same here we're not using blend (wonder if this is an issue? or we can delete it. If I comment the line, the demo doesn't change

packages/cientos/src/core/materials/meshPortalMaterial/index.vue
```diff
@@ -0,0 +1,195 @@
...
+    portalScene.matrixWorld.identity()
+  }
+  else {
+    if (props.blend === 1) { host.updateWorldMatrix(true, false) }
+    portalScene.matrixWorld.copy(host.matrixWorld)
+  }
+
+  // Keep rendering each frame: the window FBO must track animated portal contents,
+  // and the blend>=1 takeover depends on the main pass clearing the framebuffer first.
+  invalidate()
```
JaimeTorrealba, line 120: Correct me if I'm wrong but having the invalide here force the scene to re-render everytime even on "on-demand" rendering mode??? 

I undertand this is necessary right? maybe mention in the docs?

JaimeTorrealba (commented):
A couple of comments, let me know what you think.

Overall is a wonderfull job man

packages/cientos/src/core/materials/meshPortalMaterial/PortalMaterialImpl.ts
alvarosabu: They'll be reintroduced when blur/SDF lands, scoped to what that work actually needs.

packages/cientos/src/core/materials/meshPortalMaterial/index.vue
alvarosabu: I added something in the docs

JaimeTorrealba (approved):

PR #1449, opened 2026-06-24
feat(cientos): small addition to the keyboardControls with the Q key you can go down and with the E key you can go up (assuming Y+up)

JaimeTorrealba:
This imitate the behavior of UE engine and other apps like world creator


alvarosabu (approved):
Nice one. LGTM

PR #1453, opened 2026-07-10
fix(useCamera): shallow ref

posva:
found this by accident, it required me to use `toRaw()` in some places to avoid weird bugs.

I could be wrong about the fix as I don't know the rest of the codebase but since `cameras` isn't exposed, this should be fine and better for performance unless camera properties can be changed at runtime, in that case, this will break stuff


alvarosabu (approved):
Hey @posva thanks for the fix, I double-check if this would affect any of the abstactions in cientos but seems safe, and honestly, better suited.

Thanks 🙏

PR #1454, opened 2026-07-11
chore(lab): new demo world walker

JaimeTorrealba:


apps/lab/app/components/world-walker/index.global.vue
```diff
@@ -0,0 +1,47 @@
...
+const bloomRef = shallowRef()
+watch(() => bloomRef.value?.effect, (exposed) => {
+  const effect = exposed?.value ?? exposed
+  if (effect?.mipmapBlurPass) effect.mipmapBlurPass.radius = glow.radius
+}, { immediate: true })
+</script>
+
+<template>
+  <TresCanvas window-size clear-color="#111">
+    <TresPerspectiveCamera :position="[0, 8, 10]" :fov="45" :near="0.1" :far="1000" />
```
alvarosabu, line 23: The initial camera orientation doesn't allow the user to appreciate the heightmap and the environment. It feels weird that the camera follows the character as it falls.

Image

I would aim for something like this

Image

apps/lab/app/components/world-walker/fragment.glsl
alvarosabu, line 1: It's a better practice to move the shaders to it's own folder `shaders`. Lets move both fragment and vertex there and update the imports

apps/lab/app/components/world-walker/Vegetation.vue
alvarosabu, line 1: I feel the vegetation is too tall in general and opaques the character a bit, I think the primary topic of the demo is the use of heightmap for the character movement, the foliage should be secondary, try reducing the size a bit

Image

apps/lab/app/components/world-walker/Terrain.vue
alvarosabu, line 1: I thrown the poor guy on the limits of the terrain 😂 we have several options here:

1. Invisible wall, block the player movement if the direction goes towards the limit of the heightmap
2. Procedurally extend the heightmap

I would probably go for the first one since its less complex

Image

apps/lab/app/components/world-walker/index.global.vue
```diff
@@ -0,0 +1,47 @@
...
+// reaches the BloomEffect constructor. It has to be set on the mipmap pass directly.
+const bloomRef = shallowRef()
+watch(() => bloomRef.value?.effect, (exposed) => {
+  const effect = exposed?.value ?? exposed
+  if (effect?.mipmapBlurPass) effect.mipmapBlurPass.radius = glow.radius
+}, { immediate: true })
+</script>
+
+<template>
+  <TresCanvas window-size clear-color="#111">
```
alvarosabu, line 22: Please add the loading component and import it


```suggestion
<template>
  <TheLoadingScreen background="#1c1c2e" />
  <TresCanvas window-size clear-color="#111">
```

apps/lab/app/components/world-walker/Footman.vue
```diff
@@ -0,0 +1,197 @@
...
+      directionOffset = Math.PI // s
+    }
+  } else if (a.value) {
+    directionOffset = Math.PI / 2 // a
+  } else if (d.value) {
+    directionOffset = -Math.PI / 2 // d
+  }
+
+  return directionOffset
+}
```
alvarosabu, line 115: @JaimeTorrealba I present you your new best friend `atan`. The offset is a direction angle:

Collapse the whole function into:

```ts
 const getOffset = () => {
    const x = Number(a.value) - Number(d.value)
    const z = Number(w.value) - Number(s.value)
    return Math.atan2(x, z)
  }
```

Then in `moveBody`

```ts
const directionOffset = getOffset()

  // model faces -Z, so it points a half turn away from the walk
  direction
  rotateQuarternion.setFromAxisAngle(rotateAngle,
  angleYCameraDirection + directionOffset + Math.PI)
```

You don't need the extra ref nor the method

apps/lab/app/components/world-walker/Footman.vue
```diff
@@ -0,0 +1,197 @@
...
+const changeAnimation = (action: AnimationAction | undefined) => {
+  if (!action || !currentAction.value || currentAction.value === action) return
+  currentAction.value.fadeOut(fadeDuration)
+  action.reset().fadeIn(fadeDuration).play()
+  currentAction.value = action
+}
+
+// KEYS
+// useMagicKeys types its keys through an index signature, so each one comes back
+// as possibly undefined under noUncheckedIndexedAccess
```
alvarosabu, line 56: @JaimeTorrealba claude and AI tend to be over-verbose, this comment is unnecessary

apps/lab/content/experiments/world-walker.md
```diff
@@ -0,0 +1,27 @@
+---
+thumbnail: /experiments/world-walker.webp
+title: World Walker
+slug: world-walker
+author: [jaime-bboyjt]
+status: draft
+date: 2026-07-09
+featured: false
+description: A third-person character controller walking across a heightfield terrain scattered with wind-blown vegetation, under the rain. Built with @tresjs/rapier physics and cientos' Precipitation and Sky.
```
alvarosabu, line 9: From what I see there is not Precipitation or Sky used.

```suggestion
description: A third-person character controller walking across a heightfield terrain scattered with wind-blown, instanced vegetation. Built with @tresjs/rapier physics, a custom GLSL billboard shader and bloom from @tresjs/post-processing.
```

apps/lab/app/components/world-walker/Footman.vue
```diff
@@ -0,0 +1,197 @@
...
+  return directionOffset
+}
+const updateCameraTarget = (position: BodyPosition, camera: Camera, controls: Controls) => {
+  // move camera by the actual physics displacement
+  camera.position.x += position.x - prevPosition.x
+  camera.position.z += position.z - prevPosition.z
+
+  // update camera target
+  cameraTarget.set(position.x, position.y + 1, position.z)
+  controls.target = cameraTarget
```
alvarosabu, line 147: This makes controls.target and the module-scope cameraTarget the same object. `OrbitControls.reset()` does `this.target.copy(this.target0)`, which would then overwrite your scratch vector. Setting the values avoids the shared reference and cameraTarget can be deleted.


```suggestion
  controls.target.set(position.x, position.y + 1, position.z)
```

apps/lab/app/components/world-walker/Footman.vue
```diff
@@ -0,0 +1,197 @@
...
+
+// constants
+const fadeDuration = 0.2
+const velocity = 5
+const rotateAngle = new Vector3(0, 1, 0)
+const cameraTarget = new Vector3()
+const walkDirection = new Vector3()
+const rotateQuarternion = new Quaternion()
+const prevPosition = new Vector3()
+let hasPrevPosition = false
```
alvarosabu, line 38: Nit:

`velocity` holds a scalar speed, and `rotateQuarternion` is misspelled.

```ts
const speed = 5
const rotateQuaternion = new Quaternion()
```

apps/lab/app/components/world-walker/planting.ts
```diff
@@ -0,0 +1,196 @@
...
+  const min = new Vector3(Infinity, Infinity, Infinity)
+  const max = new Vector3(-Infinity, -Infinity, -Infinity)
+  let maxExtent = 0
+
+  for (let cx = 0; cx < cells; cx++) {
+    for (let cz = 0; cz < cells; cz++) {
+      for (let p = 0; p < plantsPerCell; p++) {
+        const x = originX + cx * CELL_SIZE + rand() * CELL_SIZE
+        const z = originZ + cz * CELL_SIZE + rand() * CELL_SIZE
+        const normal = sampler.normalAt(x, z)
```
alvarosabu, line 100: normalAt returns a new Vector3() per plant. With `GRASS_DENSITY = 6` that is about 270k allocations inside the awaited setup. I measured the whole planting pass at roughly 90 ms on desktop, so it is fine here, but on mobile it will be noticeable. Cheap fix: give normalAt an out-param and reuse the tmp vector 
you already have on line 55:

```ts // heightmap.ts
normalAt: (x: number, z: number, target?: Vector3) => Vector3

const normalAt = (x: number, z: number, target = new Vector3()) => {
  const nx = heightAt(x - eps, z) - heightAt(x + eps, z)
  const nz = heightAt(x, z - eps) - heightAt(x, z + eps)
  return target.set(nx, 2 * eps, nz).normalize()
}
```

```ts //here
const normal = sampler.normalAt(x, z, tmpNormal)
```

apps/lab/app/components/world-walker/index.global.vue
```diff
@@ -0,0 +1,47 @@
...
+
+const glow = {
+  intensity: 0.8,
+  threshold: 0.5,
+  smoothing: 0.35,
+  radius: 0.85,
+}
+
+// BloomPmndrs declares no `radius` prop, so binding it does nothing — the value never
+// reaches the BloomEffect constructor. It has to be set on the mipmap pass directly.
```
alvarosabu, line 13: Remove unnecessary text

alvarosabu (changes requested):
Hey @JaimeTorrealba, the demo still needs some polishing. Let me know if you have any question.

JaimeTorrealba:
I think I have solve all the issues @alvarosabu thanks for such a good feeback.

Also I add some small improvements. Let me know what you think overall

apps/lab/content/experiments/world-walker.md
```diff
@@ -0,0 +1,27 @@
+---
+thumbnail: /experiments/world-walker.webp
+title: World Walker
+slug: world-walker
+author: [jaime-bboyjt]
+status: draft
+date: 2026-07-09
```
alvarosabu, line 7: @JaimeTorrealba remember to set to published and update the date when ready ✌🏻

apps/lab/app/components/world-walker/index.global.vue
```diff
@@ -0,0 +1,25 @@
...
+<template>
+  <TheLoadingScreen background="#1c1c2e" />
+  <TresCanvas window-size clear-color="#111">
+    <!-- only frames the loading frames: Footman re-places the camera behind him once the body spawns -->
+    <TresPerspectiveCamera :position="[0, 8, 10]" :fov="45" :near="0.1" :far="1000" />
+    <!-- the horizon band of the dawn HDR, so the terrain fades into the sky rather than across it -->
+    <TresFog :color="FOG_COLOR" :near="0.1" :far="100" />
+    <Suspense>
+      <TresGroup>
+        <Environment preset="dawn" background />
```
alvarosabu, line 16: @JaimeTorrealba I think we should blur it. The background texture looks pixelated and doesnt make justice to the demo. Maybe adding fog as well would help to easy the surroundings of the mountains without vegetation.

Image

alvarosabu (changes requested):
Hi @JaimeTorrealba thanks a lot for tackling all the review comments. It looks way better al ready. Just couple of nits.

PR #1457, opened 2026-07-17
fix(rapier): improves watchers sync

Neosoulink:
## Summary

- Props sync for `RigidBody` and `Collider`
  - Better body type context injection
  - Immutable auto-collider prop updates
  - Consolidated auto-collider watchers
  - Safer unmount.
- Fix collider watchers so shape/args/pose/groups also depend on `colliderInfos` and sync when the instance appears (watcher-only material sync; Collider setters no longer get a fake `wakeUp` arg).
- Fix axis-lock conflict: `enabledRotations` / `enabledTranslations` defaults are now `undefined` instead of `[true, true, true]`, so they no longer silently call `setEnabledRotations/Translations`.


packages/rapier/src/components/colliders/Base.vue
```diff
@@ -40,7 +41,7 @@ const props = withDefaults(defineProps<Partial<ColliderProps>>(), {
 const { world } = useRapierContext()
 
 const colliderGroup = shallowRef<TresObject3D>()
-const bodyContext = inject<ShallowRef<RigidBodyContext>>('bodyContext') ?? shallowRef<RigidBodyContext>()
+const bodyContext = inject(bodyContextInjectionKey) ?? shallowRef<RigidBodyContext>()
```
alvarosabu, line 44: Appreciation

Nicely done @Neosoulink this is the proper way

packages/rapier/src/components/RigidBody.vue
```diff
@@ -41,8 +41,8 @@ const props = withDefaults(defineProps<Partial<RigidBodyProps>>(), {
   enableCcd: false,
   linvel: () => ({ x: 0, y: 0, z: 0 }),
   angvel: () => ({ x: 0, y: 0, z: 0 }),
-  enabledRotations: () => [true, true, true],
-  enabledTranslations: () => [true, true, true],
+  enabledRotations: undefined,
+  enabledTranslations: undefined,
```
alvarosabu, line 45: @Neosoulink 

Heads up with this change. The watcher in `utils/props.ts` fires as soon as instance becomes truthy, not only when the prop actually changes, so on every mount it runs with `newValue === undefined` even if the user never passed the prop, this will cascade down poorly.

Concrete fix, in [`utils/props.ts`](https://github.com/Tresjs/tres/blob/079ee33d88451812b0395a0f144e501953787b73/packages/rapier/src/utils/props.ts#L20) try bailing early in both `makePropWatcherRB` and `makePropWatcherCL`:

```ts
) => watch([() => props[toWatch], instance], ([newValue, _]) => {
  if (!instance.value || newValue === undefined) { return }
  ...
})
```

alvarosabu: `enabledRotations: undefined` in withDefaults is a no-op on its own.

alvarosabu (changes requested):
Hi @Neosoulink thanks a lot for this 🙏, I left some comments to rework but seems we are going in the correct direction

packages/rapier/src/components/RigidBody.vue
Neosoulink: Alright ✅

Neosoulink: Should I remove the following part?

```ts
  enabledRotations: undefined,
  enabledTranslations: undefined,
```

alvarosabu: yes

PR #1458, opened 2026-07-18
feat(rapier): make stepping frame-rate independent

Neosoulink:
## Summary

* Fix physics running at different speeds on different refresh rates by stepping with a fixed-timestep accumulator (and optional `timeStep="vary"`).
* Make `<Physics>` props reactive: `pause`, `timeStep`, `timeScale` (synced to `useRapier` context refs);
  * Added `timeScale` to control the simulation speed.
* Rename prop `timestep` → `timeStep` (may break some integration).
* Update demo, docs, and lab.

### Preview

https://github.com/user-attachments/assets/7682ea67-436f-4e53-b9da-f58759c58911


packages/rapier/src/components/Physics.vue
```diff
@@ -112,6 +136,33 @@ onBeforeRender(() => {
...
+    return
+  }
+
+  const fixedStep = typeof timeStep.value === 'number' ? timeStep.value : DEFAULT_TIMESTEP
+  accumulator += clampedDelta
+
+  while (accumulator >= fixedStep) {
+    stepWorld(fixedStep)
+    accumulator -= fixedStep
+  }
```
alvarosabu, line 169: Blocking:

This never terminates if `timeStep` is 0 or negative, it will freeze the tab. Please guard it

packages/rapier/src/constants/physics.ts
```diff
@@ -1 +1,8 @@
-export const GRAVITY = { x: 0, y: -9.81, z: 0 }
+/** @description Default gravity vector. */
+export const GRAVITY = { x: 0, y: -9.8, z: 0 }
```
alvarosabu:
```suggestion
export const GRAVITY = { x: 0, y: -9.81, z: 0 }
```

apps/lab/app/components/rapier-wrecking-ball/TheWreckingBall.vue
```diff
@@ -397,7 +397,7 @@ onBeforeRender(({ delta }) => {
   <primitive :object="chainMesh" />
 
   <Suspense>
-    <Physics :timestep="1 / 120">
+    <Physics :time-step="1 / 120">
```
alvarosabu, line 400: Mmm, since we are in alpha is ok, but be careful about breaking changes @Neosoulink

packages/rapier/src/components/Physics.vue
```diff
@@ -22,45 +22,69 @@ const props = withDefaults(
...
-  if (typeof props.timestep === 'number') {
-    world.value.timestep = props.timestep
-  }
+watch(timeStep, () => {
+  accumulator = 0
+})
 
-  world.value.step(eventQueue)
+watch(() => props.debug, (value) => {
+  context.isDebug.value = value
```
alvarosabu: Add the`isDebug` to L34 and replace this for


```suggestion
 isDebug.value = value
```

So its consistent to the other watchers

alvarosabu (changes requested):
Hi @Neosoulink thanks a lot for working on this, I left a couple on comments before we can merge this 🙏

packages/rapier/src/components/Physics.vue
Neosoulink: Noted ✅

apps/lab/app/components/rapier-wrecking-ball/TheWreckingBall.vue
Neosoulink: Yeah, I've added a note about it

alvarosabu (approved):
LGTM @Neosoulink thanks for taking care of the comments.

PR #1461, opened 2026-07-27
feat(rapier): add onBeforeStep callback to physics context

alvarosabu:
## Summary

Stacked on #1458, targets `feat/rapier-physics-timestep` so it reviews as a small additive diff on top of the frame-rate independent stepping work.

- Add `onBeforeStep(callback)` to the rapier context: registers a callback invoked right before every world step with the timestep about to be solved. Auto-unregisters on scope dispose, also returns a manual unregister function.
- Fire callbacks inside `stepWorld` in `<Physics>`, so with the fixed-step accumulator they run once per solved substep (0, 1, or N times per frame) in lockstep with the solver. Manual `context.step()` fires them too.
- Motivation: controllers that must apply forces per solver step, e.g. `DynamicRayCastVehicleController.updateVehicle`. With the accumulator, `onBeforeRender`-driven physics logic desyncs from the simulation.
- Docs section in `use-rapier.md` + playground demo: kinematic paddle in `/rapier/physics` driven entirely from `onBeforeStep`.

## Test plan

- [ ] Open playground `/rapier/physics`: yellow paddle spins and knocks bodies around
- [ ] `timeScale` 2 → paddle spins 2x, `0` / `pause` → halts
- [ ] Switching `timeStep` (1/60, 1/30, 1/120, 'vary') does not change perceived paddle speed


Neosoulink (approved):
Good for me ✅

PR #1479, opened 2026-08-20
fix(core): apply customRendererOptions default in TresCanvas

Jungzl:
The `options` key in withDefaults did not match any declared prop on TresCanvasProps, so the `primitivePrefix` default never reached Context.vue and `:options` was silently ignored at runtime.

Rename the default to the actual `customRendererOptions` prop and use Vue's function-form default, which is required for object types by withDefaults and prevents a shared mutable reference. This aligns the component with the documented `:custom-renderer-options` API.


alvarosabu (approved):
LGTM @Jungzl thanks a lot

Issue #1220, opened 2025-11-14
requestAnimationFrame loop keeps running in render modes on-demand, manual


vincesp:
### Describe the bug

As a developer, I expect CPU usage of my browser tab to drop to 0% if using `render-mode` `"on-demand"` or `"manual"`.<br>Actual behavior: requestAnimationFrame loop keeps running, CPU usage of 2%.

### Reproduction

[https://stackblitz.com/edit/tresjs-minimal-reproduction-4fdedjcb?file=src%2FApp.vue](<https://stackblitz.com/edit/tresjs-minimal-reproduction-4fdedjcb?file=src%2FApp.vue>)

### Steps to reproduce

Run the stackblitz and observe the console in the developer tools. It will print "advance" once and then "loop" continuously, clocking away CPU cycles needlessly.

### System Info

```shell
System:
    OS: Linux 5.0 undefined
    CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 20.19.1 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.8.2 - /usr/local/bin/npm
    pnpm: 8.15.6 - /usr/local/bin/pnpm
  npmPackages:
    @tresjs/cientos: ^5.1.0 => 5.1.0 
    @tresjs/core: ^5.1.0 => 5.1.0 
    @tresjs/leches: ^0.14.0 => 0.14.0 
    vite: ^6.0.3 => 6.0.5
```

### Used Package Manager

npm

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](<https://github.com/Tresjs/tres/blob/main/CODE_OF_CONDUCT.md>)
- [X] Read the [Contributing Guidelines](<https://github.com/Tresjs/.github/blob/main/CONTRIBUTING.md>).
- [X] Read the [docs](<https://tresjs.org/guide>).
- [X] Check that there isn't [already an issue](<https://github.com/tresjs/tres/issues>) that reports the same bug to avoid creating a duplicate.
- [X] The provided reproduction is a [minimal reproducible example](<https://stackoverflow.com/help/minimal-reproducible-example>) of the bug.

reistence:
I can reproduce this on `main` (17248d4). The render is correctly skipped, but the rAF loop keeps spinning.

[...]

PR #1488, opened 2026-09-12
fix(core): render mode resync

### LINKED ISSUE #1487: render-mode: switching from "on-demand" to "always" after idle never renders again
### Describe the bug

The render loop freezes permanently when `render-mode` is changed at runtime from `on-demand` (or `manual`) to `always`.

Verified on `main` (17248d4) and on `@tresjs/core@5.8.1`.

Root cause in `packages/core/src/composables/useRenderer/useRendererManager.ts`:

- `frames` is the internal frame budget. In `on-demand`/`manual` it drains to 0 after the pending frame is rendered.
- `notifyFrameRendered` is the only code path that re-arms the budget for `always` (`frames.value = 1`), but it runs from inside the render function, and the loop only calls the render function while `frames > 0`.
- There is no watcher on `renderMode`, so after the switch `frames` stays 0: the render function is never called again and the canvas never updates.

The same frame-budget bookkeeping also affects custom render functions: a function registered with `useLoop().render(fn)` that does not declare the `notifySuccess` parameter (e.g. `render(() => composer.render())`) cannot report completion, so in `on-demand`/`manual` the budget stays at 1 and the scene renders every frame instead of settling.

Expected behavior: switching to `always` resumes rendering every frame; `on-demand`/`manual` settle after the pending frame.

I have a fix with regression tests (8 cases) and a browser reproduction ready to submit as a PR once this issue is confirmed.

### Reproduction

https://github.com/YAKKUZA13/tres/tree/fix/core-render-mode-resync  (playground route `/core/advanced/render-mode-repro`; regression tests in `packages/core/src/composables/useRenderer/useRendererManager.test.ts`)

### Steps to reproduce

1. Run the reproduction (or paste the minimal component below into https://stackblitz.com/edit/tresjs-minimal-reproduction).
2. `on-demand` renders once and settles: the `@render` counter stays at 1.
3. Switch `:render-mode` to `always`.
4. On `main` no further renders happen: the counter stays at 1 (browser console: `BUG (renders 1 -> 1)` after 3 s). With the prepared fix the same run logs `FIXED (renders 1 -> 720)`.

<script setup lang="ts">
import { onMounted, ref } from 'vue'

const renderMode = ref<'on-demand' | 'always'>('on-demand')
let renders = 0

onMounted(async () => {
  await new Promise(resolve => setTimeout(resolve, 500))
  console.log('renders before switch:', renders)
  renderMode.value = 'always'
  await new Promise(resolve => setTimeout(resolve, 1000))
  console.log('renders after switch:', renders)
})
</script>

<template>
  <TresCanvas :render-mode="renderMode" @render="() => renders++">
    <TresPerspectiveCamera :position="[3, 3, 3]" :look-at="[0, 0, 0]" />
    <TresMesh>
      <TresBoxGeometry :args="[1, 1, 1]" />
      <TresMeshNormalMaterial />
    </TresMesh>
  </TresCanvas>
</template>

### System Info

```shell
System:
    OS: Windows 10 10.0.19045
    CPU: (32) x64 AMD Ryzen 9 8940HX with Radeon Graphics
    Memory: 17.83 GB / 31.20 GB
  Binaries:
    Node: 22.14.0 - C:\nvm4w\nodejs\node.EXE
    npm: 10.9.2 - C:\nvm4w\nodejs\npm.CMD
    pnpm: 11.6.0 - C:\nvm4w\nodejs\pnpm.CMD
  Browsers:
    Chrome: 152.0.7977.83
    Edge: Chromium (152.0.4191.53)
    Firefox: 155.0.1
  npmPackages:
    @tresjs/core: 5.8.1 (reported), 5.8.3 (reproduced on main)
    three: 0.184.0
    vue: 3.5.38
```

### Used Package Manager

pnpm

### Code of Conduct

YAKKUZA13:
## Summary

Fixes #1487: two frame-budget lifecycle bugs in `useRendererManager` that break the `on-demand`/`manual` render modes.

1. **Switching `renderMode` could permanently freeze rendering.** Once the internal frame budget (`frames`) drained to 0, switching to `always` never rendered again: `notifyFrameRendered` is the only place that re-arms the budget for `always`, but it can only run from inside the render function, and the loop only invokes the render function while `frames > 0`. There is no watcher on `renderMode`.
2. **Custom render functions without `notifySuccess` never settled.** A function registered through `useLoop().render(fn)` that does not declare the `notifySuccess` parameter (e.g. `render(() => composer.render())`) cannot report completion, so `on-demand`/`manual` kept the budget at 1 and rendered every frame.

## Changes

| Commit | Change |
| --- | --- |
| `fix(core): resync frame budget when renderMode changes` | Watcher resets `frames` on every mode change: `manual` -> `0`, `always`/`on-demand` -> `1` |
| `fix(core): settle on-demand frames when notifySuccess is omitted` | Zero-argument render functions count as one completed frame per invocation (awaits the returned promise when there is one); functions declaring `notifySuccess` keep the explicit contract |

Also:
- Documents the zero-argument behavior in the `useLoop` guide (`apps/docs`).
- Adds the first `renderMode` test coverage in `packages/core` (8 tests).

No breaking changes: behavior is unchanged for render functions that declare `notifySuccess`.

## How to test

```bash
pnpm --filter @tresjs/core test:ci
pnpm --filter @tresjs/core lint
pnpm --filter @tresjs/core build
```

| Scenario | Before | After |
| --- | --- | --- |
| `on-demand` -> `always` after the budget drained | frozen (1 render total) | resumes (4 renders over 3 frames) |
| `manual` -> `always` after the budget drained | frozen | resumes |
| `on-demand` + render function without `notifySuccess` | renders every frame | settles after 1 frame |
| `always` + render function without `notifySuccess` | renders every frame | unchanged |
| render function declaring `notifySuccess` | contract unchanged | contract unchanged |

Browser check (WebGL2, `:render-mode` toggled at runtime, 3 s window after the switch):

| Build | Idle -> `always` |
| --- | --- |
| Before (current `main`) | `renders 1 -> 1` (frozen) |
| After | `renders 1 -> 720` |

## Checklist

- [x] `pnpm --filter @tresjs/core test:ci` - 22 files, 644 passed, 3 skipped
- [x] `pnpm --filter @tresjs/core lint` - exit 0
- [x] `pnpm --filter @tresjs/core build` - exit 0
- [x] Docs updated
- [x] No breaking changes

## Notes

- The two commits are independent; happy to split them into separate PRs if separate review is preferred.
- Related to #1220: that issue tracks the rAF loop continuing to tick while idle (CPU usage). This PR does not change loop pausing; it fixes the opposite failure, where the loop stops rendering entirely after a mode switch.
- `pnpm typecheck` on the monorepo still reports 2 pre-existing errors in `useLoader/component.vue` on `main`; unchanged by this PR.

### DIFF
--- a/packages/core/src/composables/useRenderer/useRendererManager.ts
+++ b/packages/core/src/composables/useRenderer/useRendererManager.ts
@@ -335,6 +335,20 @@ export function useRendererManager(
   }
 
   const replaceRenderFunction = (fn: RenderFunction) => {
+    if (fn.length === 0) {
+      renderFunction = () => {
+        const result = (fn as () => unknown)()
+        if (result && typeof (result as Promise<unknown>).then === 'function') {
+          (result as Promise<unknown>).then(notifyFrameRendered, notifyFrameRendered)
+        }
+        else {
+          notifyFrameRendered()
+        }
+      }
+
+      return
+    }
+
     renderFunction = fn
   }
 
--- a/packages/core/src/composables/useRenderer/useRendererManager.ts
+++ b/packages/core/src/composables/useRenderer/useRendererManager.ts
@@ -269,6 +269,10 @@ export function useRendererManager(
 
   const isModeAlways = computed(() => toValue(options.renderMode) === 'always')
 
+  watch(() => toValue(options.renderMode), (renderMode) => {
+    frames.value = renderMode === 'manual' ? 0 : 1
+  })
+
   const readyEventHook = createEventHook<TresRenderer>()
   const errorEventHook = createEventHook<TresRendererError>()
   let hasTriggeredReady = false
--- a/apps/docs/content/3.api/2.composables/3.use-loop.md
+++ b/apps/docs/content/3.api/2.composables/3.use-loop.md
@@ -138,6 +138,8 @@ render((notifySuccess) => {
 
 ::warning
 **Success Callback Required**: You must call the provided callback (named `notifySuccess()` in the example above) to properly notify the render loop that the frame was completed. This is essential for the render modes (`always`, `on-demand`, `manual`) to function correctly.
+
+If the render function does not declare the `notifySuccess` parameter, each invocation is treated as a completed frame.
 ::
 
 #### Custom Rendering Examples
--- a/packages/core/src/composables/useRenderer/useRendererManager.test.ts
+++ b/packages/core/src/composables/useRenderer/useRendererManager.test.ts
@@ -1,0 +1,180 @@
+import type { RendererOptions, RenderMode } from './useRendererManager'
+import type { TresCamera } from '../../types'
+import type { ShallowRef } from 'vue'
+import { mount } from '@vue/test-utils'
+import { Scene } from 'three'
+import { computed, defineComponent, h, nextTick, reactive, ref, shallowRef } from 'vue'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import { setupMocks } from '../../utils/test/mocking'
+import { useCameraManager } from '../useCamera'
+import { useRendererManager } from './useRendererManager'
+
+const createRendererDouble = () => ({
+  domElement: { width: 800, height: 600 },
+  render: vi.fn(),
+  setSize: vi.fn(),
+  setPixelRatio: vi.fn(),
+  getPixelRatio: vi.fn(() => 1),
+  setClearColor: vi.fn(),
+  setClearAlpha: vi.fn(),
+  dispose: vi.fn(),
+  forceContextLoss: vi.fn(),
+  shadowMap: { enabled: false, type: 0 },
+  toneMapping: 0,
+  toneMappingExposure: 1,
+  outputColorSpace: '',
+})
+
+const cleanups: Array<() => void> = []
+
+const createHarness = (renderMode: RenderMode) => {
+  const sizes = {
+    width: ref(800),
+    height: ref(600),
+    pixelRatio: ref(1),
+    aspectRatio: computed(() => 800 / 600),
+  }
+  const camera = useCameraManager({ sizes })
+  camera.registerCamera({ isCamera: true, uuid: 'camera-double' } as unknown as TresCamera)
+
+  const scene = shallowRef(new Scene()) as ShallowRef<never>
+  const canvas = ref(document.createElement('canvas'))
+  const renderer = createRendererDouble()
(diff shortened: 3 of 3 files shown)


packages/core/src/composables/useRenderer/useRendererManager.ts
```diff
@@ -331,6 +335,20 @@ export function useRendererManager(
   }
 
   const replaceRenderFunction = (fn: RenderFunction) => {
+    if (fn.length === 0) {
+      renderFunction = () => {
+        const result = (fn as () => unknown)()
+        if (result && typeof (result as Promise<unknown>).then === 'function') {
+          (result as Promise<unknown>).then(notifyFrameRendered, notifyFrameRendered)
```
alvarosabu, line 342: When the render function returns a promise, the rAF loop calls it again on every tick until the first promise resolves, because frames is still 1. 

In `on-demand` this probably causes the async renders that overlap. Consider an `inFlight` flag that skips the call while a promise is pending. Please also add a test for the promise branch.

packages/core/src/composables/useRenderer/useRendererManager.ts
```diff
@@ -331,6 +335,20 @@ export function useRendererManager(
   }
 
   const replaceRenderFunction = (fn: RenderFunction) => {
+    if (fn.length === 0) {
+      renderFunction = () => {
+        const result = (fn as () => unknown)()
```
alvarosabu, line 340: mm, you probably need to add a `as` cast for the promise branch

Consider widening it, for example to (`notifySuccess: () => void) => void | Promise<unknown>`.

packages/core/src/composables/useRenderer/useRendererManager.ts
```diff
@@ -331,6 +335,20 @@ export function useRendererManager(
   }
 
   const replaceRenderFunction = (fn: RenderFunction) => {
+    if (fn.length === 0) {
```
alvarosabu, line 338: `fn.length` is 0 for rest and default parameters, so these functions do not get `notifySuccess`. Consider passing `notifyFrameRendered` to the wrapped call as well, and ignoring a second notify for the same frame.

alvarosabu (changes requested):
Hey @YAKKUZA13 thanks a lot for this PR, I tested out and it does fix the issue. Left a couple of nits. 

Let me know if you have any question

PR #1491, opened 2026-09-20
fix(core): Allow number[] for vector props

### LINKED ISSUE #696: position type is incorrect
### Description

I defined a camera and a light and used the :position attribute, but the type error was displayed in vscode; so I changed :position="[3, 3,3]" to:position="new THREE. Vector3(0, 3, 3)" results in no type error;

I don't know if it's a version problem or a package version problem.

### Suggested solution

``` vue
<script setup lang="ts">
	import * as THREE from 'three'
	import { OrbitControls, vLightHelper } from '@tresjs/cientos'

	import Texture_02 from '@/components/texture/texture_02/index.vue'
	defineOptions({
		dir_name: 'stickers',
	})
</script>

<template>
	<div class="container">
		<TresCanvas shadows clear-color="#82DBC5" preset="realistic" alpha power-preference="high-performance">
			<OrbitControls />

                        <!-- There will be a type problem here -->
			<TresPerspectiveCamera :position="[3, 3, 3]" :look-at="[0, 0, 0]" />

			<Suspense>
				<Texture_02 />
			</Suspense>

                         <!-- OK -->
			<TresDirectionalLight :intensity="1" :position="new THREE.Vector3(0, 3, 3)" v-light-helper />
		</TresCanvas>
	</div>
</template>

<style lang="scss" scoped>
	.container {
		width: 100%;
		height: 100%;
		display: flex;
		align-items: center;
		justify-content: center;
	}
</style>

```

### Alternative

At first I thought: position="new THREE.Vector3(0, 3, 3)" is a better way to understand the original parameters of threejs, but after reading the documentation, I found: position="[0,3,3]" is simpler ;But the recommended writing method in the document is: position="[0,3,3]"

image

### Additional context

image
image

### Validations

Curstantine:
This change is admittedly a bit destructive, but it solves a real headache with how Tres currently handles vector types. Right now, it expects types derived from `T['set']`, which almost always resolves to strict tuples like `[x: number, y: number, z: number]`. Almost nobody writes them like that.

Most people just pass regular arrays into props, which tsserver widens to `number[]` since it doesn't guess fixed positional indices. That immediately blows up with a type error.

Before:
image

After:
image

## Summary

- Allow numeric arrays for vector-like props on generated `<Tres*>` components.
- Fix TypeScript errors when passing values such as `number[]` or readonly tuples to props like `<TresMesh :position="position" />`.

## Changes

- Added a shared `NumericArray` type (`readonly number[]`).
- Updated `MathType`, `TresVector2`, `TresVector3`, and `TresVector4` to accept mutable arrays and readonly tuples.
- Removed redundant `Parameters<T['set']>` tuple types, since tuple values are already assignable to `NumericArray`.

closes #696


Curstantine:
I forgot to remove the `as const` in the screenshot. It does not make a difference fwiw.

packages/core/src/types/index.ts
```diff
@@ -127,15 +127,16 @@ export interface VectorCoordinates {
   z: number
 }
 
+type NumericArray = readonly number[]
+
 export type MathType<T extends MathRepresentation | THREE.Euler> = T extends THREE.Color
   ? ConstructorParameters<typeof THREE.Color> | THREE.ColorRepresentation
 
-  : T extends VectorRepresentation | THREE.Layers | THREE.Euler ? T | Parameters<T['set']> | number | VectorCoordinates : T | Parameters<T['set']>
+  : T extends VectorRepresentation | THREE.Layers | THREE.Euler ? T | NumericArray | number | VectorCoordinates : T | NumericArray
```
alvarosabu, line 135: @Curstantine This drops the Euler order from array props. `:rotation="[0, Math.PI, 0, 'YXZ']"` type-checked before and now fails, because `NumericArray` is readonly `number[]`. The runtime still supports it through `fromArray`. Please keep the tuple for `Euler`, for example `T | NumericArray | Readonly<Parameters<THREE.Euler['set']>>` in the Euler branch, and add a type test for it.

packages/core/src/types/index.ts
```diff
@@ -127,15 +127,16 @@ export interface VectorCoordinates {
...
 
 type VectorLike<VectorClass extends THREE.Vector2 | THREE.Vector3 | THREE.Vector4>
   = | VectorClass
-    | Parameters<VectorClass['set']>
-    | Readonly<Parameters<VectorClass['set']>>
+    | NumericArray
```
alvarosabu, line 139: @Curstantine With `readonly number[]`, a short array such as `[1, 2]` for position compiles. `fromArray` then writes undefined to z, and the object renders with `NaN`. This is the cost of accepting widened `number[]`.

alvarosabu (changes requested):
Hi @Curstantine thanks for stepping in and try to fix this. I ran `vue-tsc` on `:position="[3, 3, 3]"` from #696 already type-checks, so the issue no longer happens but  Widening to r`eadonly number[]` breaks `:rotation="[0, 1, 0, 'YXZ']"` and lets `:position="[1, 2]"` compile, which renders `NaN` at runtime

After considering with care I think we should re-scope this effort into the docs, mentioning that typing tuples as `satisfies TresVector3` or `as const`will do the trick. Even  casting as `Vector3Tuple` should work.

Feel free to refactor the PR accordingly or close it and open a new one for the docs. 🙏

PR #1496, opened 2026-09-22
fix(cientos): `Stage` shadow placement and props leaking into child components

sandros94:
First contribution here, and I'm pretty new to both Tres and three.js, so any feedback is more than welcome 😬

Ofc I've used some of the AI models I've access to to explore and explain stuff to me, but I did the edits myself to learn the project.

It started from `<Stage>` not placing its ground shadow at the model's bottom, and while digging a few related things came up. Each fix is its own commit, so reviewing commit by commit is probably easier:

1. shadow height: `height` was never set (lost while porting [drei's Stage](https://github.com/pmndrs/drei/blob/master/src/core/Stage.tsx)), so the shadow sat at the `<Align>` origin. It now follows the aligned content's bottom, `top`/`bottom`/`disableY` included
2. `offset` is now the full distance between content and shadow plane, instead of half of it. Content stays where `<Align>` puts it
3. the whole `shadows` object was spread into `<RandomizedLights>`, so `size` (shadow map size) was overriding the lights' frustum size and a user `bias` skipped Stage's own handling. `<Bounds>`, `<ContactShadows>` and `<AccumulativeShadows>` were also receiving Stage-only keys, with `type` even overwriting the three.js object's `.type`
4. `bias` defaults to `-0.0001` as documented, and is inverted for the accumulative lights like drei does, since that plane only receives shadows
5. `shadows: null` was crashing (`typeof null === 'object'`), now it disables shadows like `lighting` and `environment` already do
6. `MeshReflectionMaterial` had the same kind of leak, its component-only props (`resolution`, `blurSize`, `blurDepth*`, `reflectorOffset`) ended up on the material. Now stripped the same way `MeshTransmissionMaterial` already does

A few things I noticed but left out on purpose:

- `Stage` [exposes an `update()`](https://github.com/Tresjs/tres/blob/460e2e4fa9a895a8bbbe71a77d04f90910b56f1d/packages/cientos/src/core/staging/Stage.vue#L185) that does nothing. I don't know what the original intent was, so I didn't touch it
- there is no way to pass `<Bounds>` props (e.g. `duration`, `easing`) through `Stage`, neither before nor after this PR. Could be a follow-up `feat` with a `bounds` prop, like the existing `align` one. Happy to work on it if you want
- `RandomizedLights` losing its settings on `count` change and `AccumulativeShadows` never disposing its buffers are unrelated to `Stage`, in PR #1497

Everything was checked in the playground Stage demo (shadow position against the content bounds, props on the resulting three.js objects), though over software rendering as I'm currently restoring my main machine's GPU


packages/cientos/src/core/staging/Stage.vue
```diff
@@ -136,30 +136,66 @@ const lightingFillComputed: ComputedRef<[number, number, number]> = computed(()
...
-const contactShadowsComputed: ComputedRef<StageShadows | null> = computed(() => {
+const contactShadowsComputed: ComputedRef<Partial<ContactShadowsProps> | null> = computed(() => {
   if (props.shadows === true || props.shadows === 'contact') {
-    return { type: 'contact' }
+    return {}
   }
-  else if (typeof props.shadows === 'object' && props.shadows.type === 'contact') {
-    return props.shadows
+  else if (props.shadows && typeof props.shadows === 'object' && props.shadows.type === 'contact') {
+    const { type: _t, offset: _o, bias: _b, normalBias: _nb, size: _s, ...rest } = props.shadows
```
alvarosabu: Sees like for `type: 'contact'`, the RandomizedLights-only keys (`count`, `radius`, `intensity`, `ambient`, `castShadow`, `mapSize`, `near`, `position`) are not stripped. They fall through to the ContactShadows `<primitive>` and land on its group, so `position` overrides the floor placement. Worse, removing `position` again throws `Cannot assign to read only property 'position'` and breaks the Stage until reload (same on `main`). 

Please strip them here as you do in the accumulative branch. Keep `far`, since ContactShadows uses it.

alvarosabu (changes requested):
Hi @sandros94 works like charm. Small nit an we can merge

packages/cientos/src/core/staging/Stage.vue
sandros94: I've indeed missed `position` that was landing on the ContactShadows root, and removing it again threw. But while testing I've found a better approach IMO:

Instead of adding a second hand-kept list, each shadow component now only receives the props it declares (`pick(shadows, Object.keys(ContactShadows.props))`, same for `AccumulativeShadows`). This also stops ContactShadows-only keys (`blur`, `tint`, `width`, ...) from reaching `AccumulativeShadows` in accumulative mode, while `far` still reaches `ContactShadows` c5a730c

The read-only throw itself comes from core's `patchProp` [assigning `null`/`undefined` when a prop is removed](https://github.com/Tresjs/tres/blob/8d18aa4ad429fec8020b339d3d08b5f1523ab6ed/packages/core/src/core/nodeOps.ts#L476), so it isn't `Stage` specific. I'll look into a proper fix in core and since I already have #1500 on that and I prefer to to a focused commit right there instead

Longer term, I'd like to propose a nested structure for `shadows`, something like `{ type, plane: {...}, lights: {...} }` (similar to Nuxt UI's `ui` prop), so there is no key routing at all and the types can't mix. It should be additive though, keeping the current flat keys (and drei parity) working to avoid breaking changes. WDYT? I could file an issue for it to be considered later on

